PR #2, opened 2016-09-06
ELPP-1344

jamiehollern:
@nlisgo

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -1,0 +1,76 @@
+<?php
+
+use Behat\Behat\Context\SnippetAcceptingContext;
+use Behat\Gherkin\Node\PyStringNode;
+use Behat\Gherkin\Node\TableNode;
+use Drupal\DrupalExtension\Context\RawDrupalContext;
+
+/**
+ * Defines application features from the specific context.
+ */
+class FeatureContext extends RawDrupalContext implements SnippetAcceptingContext {
+
+  protected $client;
+
+  protected $verb;
+
+  protected $endpoint;
+
+  protected $headers = [];
+
+  protected $request;
+
+  protected $response;
+
+  /**
+   * Constructor.
+   */
+  public function __construct() {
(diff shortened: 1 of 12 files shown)


features/bootstrap/FeatureContext.php
```diff
@@ -0,0 +1,76 @@
...
+  protected $response;
+
+  /**
+   * Constructor.
+   */
+  public function __construct() {
+    $this->client = new \GuzzleHttp\Client(['base_uri' => 'http://127.0.0.1:8080/']);
+  }
+
+  /**
```
giorgiosironi: as discussed yesterday, these steps should be ported to plain PHPUnit (but it's still a good idea to separate them from the test classes to make the code reusable)

src/modules/jcms_rest/tests/src/SubjectsRestResourceTest.php
```diff
@@ -0,0 +1,291 @@
...
+  protected $entityTypeManager;
+
+  public function setUp() {
+    //$this->term = m::mock('Drupal\taxonomy\Entity\Term');
+    // Set the container.
+    $container = new ContainerBuilder();
+    // Entity field query.
+    $this->queryFactory = m::mock('Drupal\Core\Entity\Query\QueryFactory');
+    // Returned query (normally mock Drupal\Core\Entity\Query\Sql\Query but this
+    // causes the test to error).
```
giorgiosironi: I guess this is the "lots of mocking" you were referring to. It seems it is isolating the resource class code from the database.
My evaluation also is that this kind of test provides a low value/cost ratio, consider also that many of the mocked method calls could change without the behavior actually being incorrect (e.g. using similar method find() instead of a get(), as a result of refactoring).
So my suggestion is to take a look at the duplicated code between resources, extract that into classes (someone said composition over inheritance yesterday?) and unit test those; resource code, which to me is similar to controllers in MVC frameworks, can probably be best tested at the API level.

nlisgo:
@giorgiosironi we are going to address this feedback as part of a new piece of work. This work meets the requirements set in ELPP-1344 and the commit history will be useful for when we reintroduce behat at a later stage.

PR #3, opened 2016-09-15
ELPP-1344

jamiehollern:
@nlisgo

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,10 @@
     "mockery/mockery": "~0.9",
     "satooshi/php-coveralls": "^1.0",
     "phpunit/phpcov": "2.*",
-    "bossa/phpspec2-expect": "~1.0"
+    "bossa/phpspec2-expect": "~1.0",
+    "puli/cli": "^1.0",
+    "elife/api": "dev-master",
+    "elife/api-validator": "dev-master"
   },
   "minimum-stability": "beta",
   "prefer-stable": true,
--- a/src/modules/jcms_rest/src/EventSubscriber/ExceptionElifeJsonSubscriber.php
+++ b/src/modules/jcms_rest/src/EventSubscriber/ExceptionElifeJsonSubscriber.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Drupal\jcms_rest\EventSubscriber;
+
+use Drupal\Core\EventSubscriber\ExceptionJsonSubscriber;
+
+/**
+ * Handle eLife JSON exceptions the same as JSON exceptions.
+ */
+class ExceptionElifeJsonSubscriber extends ExceptionJsonSubscriber {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getHandledFormats() {
+    return ['elife_json'];
+  }
+
+}
--- a/src/modules/jcms_rest/tests/src/EndpointValidatorTest.php
+++ b/src/modules/jcms_rest/tests/src/EndpointValidatorTest.php
@@ -1,0 +1,66 @@
+<?php
+namespace Drupal\Tests\jcms_rest\Unit;
+
+use Drupal\Tests\UnitTestCase;
+use GuzzleHttp\Client;
+use GuzzleHttp\Psr7\Request;
+use eLife\ApiValidator\MessageValidator\JsonMessageValidator;
+use eLife\ApiValidator\SchemaFinder\PuliSchemaFinder;
+use Webmozart\Json\JsonDecoder;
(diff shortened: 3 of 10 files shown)


src/modules/jcms_rest/tests/src/EndpointValidatorTest.php
```diff
@@ -0,0 +1,66 @@
...
+    // "Puli\GeneratedPuliFactory" is not found by the autoloader. In this case,
+    // we load it in manually.
+    $this->projectRoot = realpath(__DIR__ . '/../../../../..');
+    if (!class_exists('Puli\GeneratedPuliFactory')) {
+      if (file_exists($this->projectRoot . '/.puli/GeneratedPuliFactory.php')) {
+        require_once($this->projectRoot . '/.puli/GeneratedPuliFactory.php');
+      }
+    }
+    $this->client = new Client([
+      'base_uri' => 'http://127.0.0.1:8080/',
```
giorgiosironi: if this is making real HTTP requests, it will be on port 80 or 443 (for https) in testing environments

src/modules/jcms_rest/tests/src/EndpointValidatorTest.php
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * @test
+   * @expectedException \eLife\ApiValidator\Exception\SchemaNotFound
+   */
+  public function testSubjectsEndpoint() {
+    $this->markTestIncomplete(
+      'This test has not been implemented yet.'
+    );
+    $request = new Request('GET', '/subjects', [
+      'Content-Type' => 'application/vnd.elife.annual-report-list+json;version=1',
```
giorgiosironi: I'm not sure what the api-raml says, but GET requests should have no `Content-Type` header as they have no body like POST ones. The response, however, should have this precise `Content-Type` header.

src/modules/jcms_rest/src/EventSubscriber/ExceptionElifeJsonSubscriber.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Drupal\jcms_rest\EventSubscriber;
+
+use Drupal\Core\EventSubscriber\ExceptionJsonSubscriber;
+
+/**
+ * Handle eLife JSON exceptions the same as JSON exceptions.
+ */
+class ExceptionElifeJsonSubscriber extends ExceptionJsonSubscriber {
```
giorgiosironi: what does this class do? If there is an exception in the resource class, where can I see it? (in the response or in some log)

composer.json
```diff
@@ -54,7 +54,10 @@
...
     "satooshi/php-coveralls": "^1.0",
     "phpunit/phpcov": "2.*",
-    "bossa/phpspec2-expect": "~1.0"
+    "bossa/phpspec2-expect": "~1.0",
+    "puli/cli": "^1.0",
+    "elife/api": "dev-master",
+    "elife/api-validator": "dev-master"
   },
   "minimum-stability": "beta",
   "prefer-stable": true,
```
giorgiosironi: originally @nlisgo decided not to commit `composer.lock` due to some issues with the build process where some steps would not be executed when using just `composer install` with a known `composer.lock`. Are these reasons still present?

src/modules/jcms_rest/tests/src/EndpointValidatorTest.php
jamiehollern: This was simply because in the Travis build Nathan was already spinning up a Drupal instance on this port, so I just used that. In terms of ports, the RAML schema fails against HTTP anyway so this will need updated.

src/modules/jcms_rest/tests/src/EndpointValidatorTest.php
jamiehollern: No worries, will update.

src/modules/jcms_rest/src/EventSubscriber/ExceptionElifeJsonSubscriber.php
jamiehollern: This basically tells Drupal that if there's an exception when using the custom JSON format then handle it in the same way as the normal JSON exceptions are handled. As for the second part of the question, I'm unsure. They'd be reflected in the response via a response code but I'll need to look further into this.

jamiehollern:
Based on discussion with @nlisgo I'm closing this and will open a new one later.

composer.json
nlisgo: @jamiehollern we should commit the composer.lock as well.

PR #4, opened 2016-09-19
ELPP-1344

jamiehollern:
@nlisgo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ scripts/legacy_cms.sql.gz
 scripts/legacy_cms_files/
 vendor/
 web/
+build
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,3 @@
+coverage_clover: build/logs/clover.xml
+json_path: build/logs/coveralls-upload.json
+service_name: travis-ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: php
 php:
-  - 5.6
   - 7.0
 
 sudo: false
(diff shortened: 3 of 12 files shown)


src/modules/jcms_rest/src/JcmsRestServiceProvider.php
```diff
@@ -0,0 +1,22 @@
...
+ * Adds application/vnd.elife.annual-report-list+json as a known format.
+ */
+class JcmsRestServiceProvider implements ServiceModifierInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function alter(ContainerBuilder $container) {
+    if ($container->has('http_middleware.negotiation') && is_a($container->getDefinition('http_middleware.negotiation')->getClass(), '\Drupal\Core\StackMiddleware\NegotiationMiddleware', TRUE)) {
+      $container->getDefinition('http_middleware.negotiation')->addMethodCall('registerFormat', ['jcms_json', ['application/vnd.elife.annual-report-list+json']]);
```
giorgiosironi: there is also a `;version=1` suffix in the `Accept` header, which is sent by journal, so that it's actually `application/vnd.elife.annual-report-list+json;version=1`. However, this class seems to be not particularly tied to any of the APIs, while this header changes depending on which API is being called.

jamiehollern: The content type is indeed `application/vnd.elife.annual-report-list+json` and anything after that is a parameter. That means we can accept various different versions and aren't tied to a single one.

.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 .idea/
 .vagrant/
+.puli
```
nlisgo: `.puli/`

PR #7, opened 2016-10-04
ELPP-897 Endpoint can be validated by the RAML schema.

jamiehollern:
@nlisgo @giorgiosironi @thewilkybarkid


.travis.yml
```diff
@@ -13,6 +13,8 @@ services:
   - mysql
 
 before_install:
+  - curl https://puli.io/installer | php
+#  - mv puli.phar /usr/local/bin/puli
```
nlisgo: Do we need this line anymore?

Jenkinsfile
```diff
@@ -4,13 +4,10 @@ elifePipeline {
...
-    // def testArtifact = "${env.BUILD_TAG}.junit.xml"
-    builderDeployRevision 'journal-cms--ci', commit
-    // only smoke tests will run inside this step:
-    builderProjectTests 'journal-cms--ci', '/srv/journal-cms'
-    // builderTestArtifact testArtifact, 'journal-cms--ci', '/srv/journal-cms/build/phpunit.xml'
-    // elifeVerifyJunitXml testArtifact
+    lock('journal-cms--ci') {
+        builderDeployRevision 'journal-cms--ci', commit
+        builderProjectTests 'journal-cms--ci', '/srv/journal-cms', ['build/phpunit.xml']
+    }
```
nlisgo: Has @giorgiosironi requested these changes?

src/modules/jcms_rest/tests/src/EndpointValidatorTest.php
```diff
@@ -35,30 +37,29 @@ public function setUp() {
         require_once($this->projectRoot . '/.puli/GeneratedPuliFactory.php');
       }
     }
+    // Setup Puli.
+    $this->resourceRepository = (new GeneratedPuliFactory)->createRepository();
+    /*$factoryClass = PULI_FACTORY_CLASS;
+    $this->resourceRepository = (new $factoryClass())->createRepository();*/
```
nlisgo: Why do we still need this code?

.travis.yml
jamiehollern: No, I'll remove it.

src/modules/jcms_rest/tests/src/EndpointValidatorTest.php
jamiehollern: We don't need it at the moment but it was left in so we could use it later if the situation changes as the current implementation is hacky. Would you like it removed?

nlisgo: Just remove it. Would need more information to be valuable.

PR #8, opened 2016-10-06
Apply patch to config_installer to address PHP warning

nlisgo:


giorgiosironi:
For some reason the `composer update` of this seems to trigger this error:
`Fatal error: Uncaught Error: Call to undefined method Puli\ComposerPlugin\PuliPluginImpl::preAutoloadDump() in /srv/journal-cms/vendor/puli/composer-plugin/src/PuliPlugin.php:83`
while the build on `develop` is green. In other projects we use the older `composer1.0` instead of the current Composer, we should probably do the same here, I'll update the formula.

giorgiosironi:
Now that `puli` and `composer1.0` are present, I'm testing this with a deploy from scratch, that triggers `composer update` commands.

giorgiosironi:
Confirmed `composer update` works again, build is green.

PR #9, opened 2016-10-11
ELPP-897 Updated the branch so that an endpoint is validated by the R…

jamiehollern:
…AML schema.

Allow migrate-import of json fixtures by file rather than http (#1)

Preparing PHPUnit tests to run inside Alfred build

Apply patch to config_installer to address PHP warning

ELPP-897 Updates.


giorgiosironi:
This appears to have broken the build (https://ci--alfred.elifesciences.org/job/test-journal-cms/48/console). In fact, the PR said `1 of 2 checks passed` as the one on Alfred did not (https://ci--alfred.elifesciences.org/job/pull-requests-projects/job/journal-cms/job/PR-9/2/console)

giorgiosironi:
Investigating on the reason, the tests seem to contact localhost on port `8080`, but in all the EC2 infrastructure Nginx runs on the standard HTTP port `80`. How do we go about making this uniform across environment or configuring it somewhere?

jamiehollern:
We could change the port on our Travis build environment to 80 but I don't know if that's possible.

giorgiosironi:
You would need to change `sudo: false` to `sudo: true` in order do that, but it's possible.

PR #11, opened 2016-10-13
Refactor dependencies

nlisgo:
This pull request does not address the issue of the tests not being able to be run locally or on Albert (our Jenkins instance).

But some proposals of how to reduce the dependencies on the host machine, the only real dependency within the composer.json needs to be drupal-vm.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,39 +1,34 @@
 language: php
+
 php:
   - 7.0
 
-sudo: false
-
-env:
-  global:
-    - SIMPLETEST_DB=mysql://root:@localhost/journal_cms
-    - SIMPLETEST_BASE_URL="http://127.0.0.1:8080"
-
-services:
-  - mysql
+mysql:
+  database: journal_cms
+  username: root
+  encoding: utf8
 
 before_install:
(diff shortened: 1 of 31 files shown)


.travis.yml
```diff
@@ -1,39 +1,36 @@
...
   - echo 'sendmail_path = /bin/true' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - composer --verbose self-update
   - composer --version
 
 install:
-  - mysql --execute='CREATE DATABASE journal_cms;'
   # We require coveralls like this as it won't work straight from composer.json.
   - composer require satooshi/php-coveralls
   - composer --verbose install
-  - composer drupal-scaffold
```
jamiehollern: Should this be removed? I wasn't sure if Composer ran the scaffold script if there was a lock file present.

giorgiosironi: Does that mean we could do `composer install` too in https://github.com/elifesciences/journal-cms-formula/blob/master/salt/journal-cms/init.sls#L51 instead of `composer update`?

PR #16, opened 2016-10-17
Amend basic_html configuration and filter content that may contain HTML on migration

nlisgo:
@jamiehollern available for review.


src/modules/jcms_migrate/src/Plugin/migrate/process/JCMSSplitContent.php
```diff
@@ -253,6 +243,7 @@ public function DOMnodeInnerHTML($element) {
       foreach ($element as $child) {
         $html = $child->ownerDocument->saveHTML($child);
         $html = str_replace('&nbsp;', ' ', $html);
+        $html = check_markup($html, 'basic_html');
```
nlisgo: @jamiehollern The check_markup function is not available when the unit tests are run, what approach should I take to resolve this issue?

jamiehollern: @nlisgo There are a couple of ways you can do it: https://www.drupal.org/docs/8/phpunit/unit-testing-more-complicated-drupal-classes

PR #25, opened 2016-11-03
Create event content type and rest resource

nlisgo:


giorgiosironi:
The build fails with `Unable to install the geofield module since it does not exist.`

Is it at the PHP/Composer level or is there any Ubuntu package you need to be added to the EC2 instances?

nlisgo:
@giorgiosironi I will address this. It was a module that I installed at one stage and have not removed all references to it.

giorgiosironi:
To build this I added https://github.com/elifesciences/journal-cms-formula/commit/235e458ea88bd6a100b5c1ee0ebcfc6148566b7b which does the same thing as the Travis build.

It is unclear which version of Puli this project uses?
- the `curl https://puli.io/installer | php` installs `1.0.0-beta10` globally
- `composer.json` requires `puli/cli` as `1.0.x-dev`
- `composer.lock` requires `puli/cli` as `dev-master` and in fact Composer complains about it being not up to date with `composer.json`

nlisgo:
@giorgiosironi I will look into the version conflicts in a future piece of work. Thanks for clearing the break on alfred.

PR #27, opened 2016-11-10
ELPP-1455 Article content type and SQS integration.

jamiehollern:
@nlisgo @giorgiosironi 

Todo: Add test coverage and document new settings.

### DIFF
--- a/src/modules/jcms_notifications/jcms_notifications.module
+++ b/src/modules/jcms_notifications/jcms_notifications.module
@@ -1,0 +1,23 @@
+<?php
+
+/**
+ * @file
+ * Contains jcms_notifications.module.
+ */
+
+use Drupal\Core\Routing\RouteMatchInterface;
+
+/**
+ * Implements hook_help().
+ */
+function jcms_notifications_help($route_name, RouteMatchInterface $route_match) {
+  switch ($route_name) {
+    // Main module help for the jcms_notifications module.
+    case 'help.page.jcms_notifications':
+      $output = '';
+      $output .= '<h3>' . t('About') . '</h3>';
+      $output .= '<p>' . t('Integrates with Amazon SQS and Amazon SNS.') . '</p>';
+      return $output;
+    default:
+  }
+}
--- a/src/modules/jcms_notifications/src/FetchArticleService.php
+++ b/src/modules/jcms_notifications/src/FetchArticleService.php
@@ -1,0 +1,110 @@
+<?php
+
+namespace Drupal\jcms_notifications;
+
+use Drupal\Core\Site\Settings;
+use GuzzleHttp\Client;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Class FetchArticleService.
+ *
+ * @package Drupal\jcms_notifications
+ */
+class FetchArticleService {
+
+  /**
+   * @var \GuzzleHttp\Client
+   */
+  protected $client;
+
+  protected $endpoint;
+
+  /**
+   * FetchArticleService constructor.
+   *
+   * @param \GuzzleHttp\Client $client
+   */
+  public function __construct(Client $client) {
+    $this->client = $client;
+    $this->endpoint = Settings::get('jcms_articles_endpoint');
+  }
+
+  /**
+   * Gets the articles by the IDs from SQS.
+   *
+   * @param string $id
+   *
+   * @return array
+   */
+  public function getArticle(string $id) {
+    $article = [];
+    if ($id) {
+      $unpublished = $this->makeRequest($id, TRUE);
+      if ($unpublished->getStatusCode() == 404) {
+        $article = [
+          'id' => $id,
+          'action' => 'del',
+        ];
+      }
+      else {
+        $published = $this->makeRequest($id);
+        $article = [
+          'id' => $id,
+          'action' => 'write',
+          'data' => [
+            'unpublished' => $this->getSnippet($unpublished->getBody()->getContents()),
+            'published' => $published->getBody()->getContents() ? $this->getSnippet($published->getBody()->getContents()): '',
+          ],
+        ];
+      }
+    }
+    return $article;
+  }
+
+  /**
+   * Helper method to make the requests to get the articles.
+   *
+   * @param string $id
+   * @param bool $unpublished
+   *
+   * @return \Psr\Http\Message\ResponseInterface
+   * @throws \TypeError
+   */
+  protected function makeRequest(string $id, $unpublished = FALSE) {
+    $auth_key = 'jcms_article_auth_' . ($unpublished ? 'unpublished' : 'published');
+    $options = ['auth' => Settings::get($auth_key)];
+    $url = $this->formatUrl($id, $this->endpoint);
+    $response = $this->client->get($url, $options);
+    if ($response instanceof ResponseInterface) {
+      return $response;
+    }
+    else {
+      throw new \TypeError('Network connection interrupted on request.');
+    }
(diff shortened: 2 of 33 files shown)


src/modules/jcms_notifications/jcms_notifications.module
```diff
@@ -0,0 +1,38 @@
+<?php
+
+/**
+ * @file
+ * Contains jcms_notifications.module..
```
jamiehollern: Ah! This is silly drupal:generate code. Will update.

giorgiosironi:
The build failed during the site install with `Unable to install the jsonb module since it does not exist`, check it out later.

src/modules/jcms_notifications/src/JCMSQueueService.php
```diff
@@ -0,0 +1,95 @@
...
+      $message = $this->sqsClient->receiveMessage([
+        'QueueUrl' => $queue['QueueUrl'],
+        'VisibilityTimeout' => 60,
+        'WaitTimeSeconds' => 20,
+      ]);
+      $sqs_message = $message->get('Messages');
+      if ($sqs_message !== NULL) {
+        $sqs_message = array_shift($sqs_message);
+        $messages[] = $sqs_message;
+        $this->sqsClient->deleteMessage([
```
giorgiosironi: the `deleteMessage` that acknowledges delivery seems to be located too soon; this call permanently removes the message from the queue, therefore the pattern is:
```
while (true) {
    $message = $sqsClient->receiveMessage();
    // make HTTP calls, insert in database, commit transaction, do whatever you need to do...
    $sqsClient->deleteMessage(...);
```

src/modules/jcms_notifications/jcms_notifications.module
```diff
@@ -0,0 +1,38 @@
...
+      return $output;
+
+    default:
+  }
+}
+
+/**
+ * Implements hook_cron().
+ */
+function jcms_notifications_cron() {
```
giorgiosironi: how is this triggered from outside? Some `drush` command?

src/modules/jcms_notifications/src/JCMSQueueService.php
```diff
@@ -0,0 +1,95 @@
...
+   * Gets the article data from SQS.
+   *
+   * @return array
+   */
+  public function getArticleData() {
+    $queue = $this->sqsClient->getQueueUrl([
+      'QueueName' => $this->queueName,
+    ]);
+    $messages = [];
+    while (TRUE) {
```
giorgiosironi: at this level I wouldn't loop as this will stay inside the while as long as it consume at least one message. It is true we may want a loop to quickly go through a large queue, but this is usually put at the highest level, in our case `jcms_notifications_cron()`

giorgiosironi: to be clear: it would be fine for this function to consume just one message

src/modules/jcms_notifications/jcms_notifications.module
jamiehollern: It can be run with a `drush` command or alternatively just set up the old fashioned way as a straight cron job. See https://www.drupal.org/docs/7/setting-up-cron-for-drupal/configuring-cron-jobs-using-the-cron-command

jamiehollern:
@giorgiosironi The build failed because the composer.lock file was wrong. Fixed it now.

src/modules/jcms_notifications/jcms_notifications.module
giorgiosironi: I second `drush` and would avoid exposing the cron over an URL, this needs to be a cli process

nlisgo:
@jamiehollern do we understand what the issue is with the build on alfred? Is it because of the MySQL version?

jamiehollern:
@nlisgo It looks like it to me. It's the same errors I had on Travis before getting MySQL upgraded there. CC @giorgiosironi.

nlisgo:
I have created a PR to your feature branch addressing a few nit picking issues:

https://github.com/digirati-co-uk/journal-cms/pull/2

src/modules/jcms_notifications/src/FetchArticleService.php
```diff
@@ -0,0 +1,108 @@
...
+    $this->client = $client;
+    $this->endpoint = Settings::get('jcms_articles_endpoint');
+  }
+
+  /**
+   * Gets the articles by the IDs from SQS.
+   *
+   * @return array
+   */
+  public function getArticle(int $id) {
```
nlisgo: Am I correct in thinking that `$id` is the queue ID rather than the article ID?

giorgiosironi:
Yes those look like MySQL errors, taking out http://jira.elifesciences.org:8080/browse/ELPP-1689 for today

src/modules/jcms_notifications/src/FetchArticleService.php
jamiehollern: No, this is the article ID.

nlisgo: The article ID is not an integer, it's a string.

src/modules/jcms_notifications/src/FetchArticleService.php
```diff
@@ -0,0 +1,108 @@
...
+        ];
+      }
+    }
+    return $article;
+  }
+
+  /**
+   * Helper method to make the requests to get the articles.
+   *
+   * @param int $id
```
nlisgo: `@param string $id`

src/modules/jcms_notifications/src/FetchArticleService.php
```diff
@@ -0,0 +1,108 @@
...
+   * @param \GuzzleHttp\Client $client
+   */
+  public function __construct(Client $client) {
+    $this->client = $client;
+    $this->endpoint = Settings::get('jcms_articles_endpoint');
+  }
+
+  /**
+   * Gets the articles by the IDs from SQS.
+   *
```
nlisgo: `@param string $id`

src/modules/jcms_notifications/src/FetchArticleService.php
```diff
@@ -0,0 +1,108 @@
...
+    }
+    else {
+      throw new \TypeError('Network connection interrupted on request.');
+    }
+  }
+
+  /**
+   * Helper method to format an URL with a correct ID.
+   *
+   * @param int $id
```
nlisgo: `@param string $id`

src/modules/jcms_notifications/src/FetchArticleService.php
```diff
@@ -0,0 +1,108 @@
...
+   * @param string $url
+   *
+   * @return string
+   */
+  protected function formatUrl(int $id, string $url) : string {
```
nlisgo: `string $id`

giorgiosironi:
http://jira.elifesciences.org:8080/browse/ELPP-1689 is necessary for testing and merging.

giorgiosironi:
MySQL 5.7 is there, but `composer install` failed with `Source directory web/modules/contrib/restui has uncommitted changes.` so I'm removing the folder and retrying from scratch.

giorgiosironi:
Provisioned correctly, now tests fail with the error
```
Error: Class 'Puli\GeneratedPuliFactory' not found
```
which I remember we already encountered. There was some problematic combination of Composer and Puli versions?

giorgiosironi:
Now green, for reference `puli.phar` has to be in `/srv/journal-cms`.

PR #32, opened 2016-11-21
Create press-packages endpoint

nlisgo:


jamiehollern:
Looks fine.

jamiehollern:
As of Drupal 8.2 composer support is native so you can just include it in a composer.json in the module itself.

composer.json
```diff
@@ -50,7 +49,8 @@
     "justinrainbow/json-schema": "^2.0",
     "drupal/field_permissions": "~8.1.0-beta1",
     "drupal/address": "^8.1@RC",
-    "aws/aws-sdk-php": "~3.0"
+    "aws/aws-sdk-php": "~3.0",
+    "giggsey/libphonenumber-for-php": "^7.7"
```
nlisgo: I will overlook this for now and perhaps it can be addressed in a future PR. I had a quick go and it didn't work.

jamiehollern: According to [Add a composer.json file](https://www.drupal.org/docs/8/creating-custom-modules/add-a-composerjson-file), "You may optionally define external dependencies for your module in composer.json. **Drupal core will not automatically discover or manage these dependencies.**"

The module must be defined properly in the composer.json "which at a minimum defines the name, description and type of your module" and the module must also be installed using composer.

```
{
    "name": "drupal/example",
    "description": "This is an example composer.json for example module.",
    "type": "drupal-module",
    "require": {
        "mobiledetect/mobiledetectlib": "~2.8"
    }
}
```

PR #33, opened 2016-11-21
Consistently referring to JcmsNotAcceptableHttpException

giorgiosironi:
Was referred to as `JCMSNotAcceptableHttpException` which is probably causing a class not found error on Ubuntu servers when `Accept` headers are incorrect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ before_install:
   - sh $TRAVIS_BUILD_DIR/scripts/travis/mysql-reset-root-password.sh
   - curl https://puli.io/installer | php
   - echo 'sendmail_path = /bin/true' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
+  - echo 'xdebug.max_nesting_level = 512' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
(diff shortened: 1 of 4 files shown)


giorgiosironi:
The OS X filesystem is sometimes case-insensitive so it may not reproduce this:
```
[21-Nov-2016 08:56:15 America/Chicago] Error: Class 'Drupal\jcms_rest\Exception\JCMSNotAcceptableHttpException' not found in /srv/journal-cms/src/modules/jcms_rest/src/Routing/JcmsRestResourceRouteFilter.php on line 66 #0 /srv/journal-cms/web/core/lib/Drupal/Core/Routing/LazyRouteFilter.php(95): Drupal\jcms_rest\Routing\JcmsRestResourceRouteFilter->filter(Object(Symfony\Component\Routing\RouteCollection), Object(Symfony\Component\HttpFoundation\Request))
```

giorgiosironi:
Not sure how the change is related, but there seem to be a default limit set by `xdebug.max_nesting_level`. On Alfred, which does not install xdebug, PHPUnit runs to completion so I'll try to raise it in `.travis.yml`.

giorgiosironi:
Ok, now works on Travis CI too. Merge when satisfied.

PR #40, opened 2016-11-25
Create covers content type and migrate legacy covers

nlisgo:
@jamiehollern

### DIFF
--- a/src/modules/jcms_migrate/src/Plugin/migrate/process/JCMSCollectionContent.php
+++ b/src/modules/jcms_migrate/src/Plugin/migrate/process/JCMSCollectionContent.php
@@ -20,6 +20,12 @@ class JCMSCollectionContent extends AbstractJCMSContainerFactoryPlugin {
   public function transform($value, MigrateExecutableInterface $migrate_executable, Row $row, $destination_property) {
     if (!empty($value)) {
       if (!isset($this->configuration['multiple']) || $this->configuration['multiple'] === FALSE) {
+        if (is_string($value)) {
+          if (strpos($value, '{') === FALSE) {
+            $value = '{' . $value . '}';
+          }
+          $value = json_decode($value, TRUE);
+        }
         return $this->processItemValue($value['type'], $value['source'], $migrate_executable, $row, $destination_property);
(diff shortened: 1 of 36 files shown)


src/modules/jcms_migrate/src/Plugin/migrate/process/JCMSCollectionContent.php
```diff
@@ -20,6 +20,12 @@ class JCMSCollectionContent extends AbstractJCMSContainerFactoryPlugin {
   public function transform($value, MigrateExecutableInterface $migrate_executable, Row $row, $destination_property) {
     if (!empty($value)) {
       if (!isset($this->configuration['multiple']) || $this->configuration['multiple'] === FALSE) {
+        if (is_string($value)) {
+          if (strpos($value, '{') === FALSE) {
```
jamiehollern: Is this to check if it's JSON? If so perhaps a more elegant check would be worth it here?

src/modules/jcms_migrate/src/Plugin/migrate/process/JCMSImage.php
```diff
@@ -35,7 +34,7 @@ public function transform($value, MigrateExecutableInterface $migrate_executable
         $file->save();
         return [
           'target_id' => $file->id(),
-          'alt' => $alt,
+          'alt' => !empty($alt) ? $alt : '',
```
jamiehollern: Just a heads up, this can be done as `'alt' => $alt ?: '',`

src/modules/jcms_migrate/src/Plugin/migrate/process/JCMSCollectionContent.php
nlisgo: The expectation if it is a string is that it is json but when preparing the expression in the sql query it was stripping out the curly brackets. So, if the string doesn't start with `{` and it is a string then I wrap it in curly brackets.

The expression is here:
https://github.com/elifesciences/journal-cms/pull/40/files#diff-715cd3ad4888a4dc0889bb60b0cf12a3R32

PR #41, opened 2016-12-01
Some basic stdout logs for what is happening when ingesting SQS messages

giorgiosironi:


giorgiosironi:
Ok, ready for review @jamiehollern

PR #42, opened 2016-12-01
Add covers queue and API endpoint

nlisgo:
We are still awaiting covers to be defined in API raml. When it is introduced we may need to do some work to ensure that the response is as expected. This work is just preliminary so that we can verify that the endpoint is active and returns appropriate data.

@jamiehollern


sync/system.date.yml
```diff
@@ -1,8 +1,8 @@
 country:
-  default: ''
-first_day: 0
+  default: GB
+first_day: 1
 timezone:
-  default: America/Chicago
+  default: UTC
```
giorgiosironi: :+1:

src/modules/jcms_rest/src/Plugin/rest/resource/CoverListRestResource.php
```diff
@@ -0,0 +1,111 @@
...
+   * @param \Drupal\Core\Entity\EntityInterface $node
+   *
+   * @return array
+   */
+  public function getItem(EntityInterface $node) {
+    /* @var Node $node */
+    /* @var Node $related */
+    $related = $node->get('field_cover_content')->first()->get('entity')->getTarget()->getValue();
+    $rest_resource = [
+      'blog_article' => new BlogArticleListRestResource([], 'blog_article_list_rest_resource', [], $this->serializerFormats, $this->logger),
```
jamiehollern: This is clever, and shows that we can probably retrofit DI into our rest resources when we have time.

sync/system.date.yml
jamiehollern: 🍾

PR #46, opened 2016-12-13
ELPP-1815 Re-added the creation logic in the article CRUD service.

jamiehollern:
@nlisgo @giorgiosironi

### DIFF
--- a/src/modules/jcms_notifications/src/ArticleCrudService.php
+++ b/src/modules/jcms_notifications/src/ArticleCrudService.php
@@ -51,6 +51,9 @@ class ArticleCrudService {
         $node = $this->updateArticle($article);
       }
     }
+    else {
+      $node = $this->createArticle($article);
+    }
     return $node;
   }
 
@@ -88,7 +91,6 @@ class ArticleCrudService {
         ],
       ];
     }
-
     $node->save();
     return $node;
   }


giorgiosironi:
Is there any test you can add at the project level for this code change (test that fails without it but passes afterwards)?

jamiehollern:
Yeah, I'll get some in. This class needs tests anyway.

jamiehollern:
@giorgiosironi The last commit adds test coverage and will automatically fail if the option of creating new articles does not exist.

src/modules/jcms_notifications/tests/src/Unit/ArticleCrudServiceTest.php
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  /**
+   * @test
+   * @covers \Drupal\jcms_notifications\ArticleCrudService::crudArticle()
+   * @covers \Drupal\jcms_notifications\ArticleCrudService::nodeExists()
+   * @covers \Drupal\jcms_notifications\ArticleCrudService::updateArticle()
+   */
+  public function testCrudArticleUpdate() {
+    $node = m::mock('\Drupal\node\Entity\Node');
```
giorgiosironi: tests that mock ORM calls are going to haunt you in the future as the set of dozens of calls change without the behavior being different. The stable way to test this is at the integration level, actually writing in the database a random article id and checking that is is there. Given that journal-cms has no knowledge of articles before the SQS message coming in, the setup necessary for such a test should be minimal?

nlisgo: @giorgiosironi is your recommendation that this test should be dropped. The integration tests that you refer to, should they be add in this project or another?

jamiehollern: @nlisgo I've already started Kernel Tests for this but got stuck on some AWS config. I'll run you through the approach once I'm finished what I'm currently doing.

giorgiosironi: recommending to drop this as a unit tests (mocking a dozen of objects to make it work) and implement it as an integration test inside this project instead (using the real MySQL database underneath by driving method calls to an ArticleCrudService object instead, everything in the single PHPUnit process)

giorgiosironi:
Trying to realistically merge this: drop the test, keep the createArticle() call change, we get it in end2end and test everything together.

jamiehollern:
@giorgiosironi Unit tests removed.

giorgiosironi:
Oddly we had a build failure for a JSON syntax error, what are these tests doing?
https://ci--alfred.elifesciences.org/job/pull-requests-projects/job/journal-cms/job/PR-46/3/console

jamiehollern:
@giorgiosironi The tests make a call to the various endpoints with the correct content types and validates the JSON schema and response codes. It seems there may be an issue with the JSON being returned for the endpoints `/subjects/{id}`, `labs-experiments/{id}` and `community/{id}`.

PR #49, opened 2016-12-20
Upgrade to php 7.1

nlisgo:


giorgiosironi:
Ondrej's PPA seem to have this, do we want it in general for our PHP projects?

thewilkybarkid:
@giorgiosironi: Yep, ELPP-1921.

nlisgo:
@giorgiosironi this is to ensure that journal-cms is ready on php 7.1 by introducing it for development and on TravisCI. Where is the appropriate place to change the PHP version for the deployment instance? journal-cms-formula?

giorgiosironi:
https://github.com/elifesciences/builder-base-formula/blob/master/elife/php7.sls#L22 which is intentionally shared by all 2.0 PHP projects, as we have shared libraries between them. We will push out the update for all of them, but probably the rollout will need a different `php71.sls` file to do one at a time.

PR #50, opened 2016-12-21
ELPP-1759 Article ingestion via Drush.

jamiehollern:
@nlisgo @giorgiosironi

### DIFF
--- a/config/drupal-vm.settings.php
+++ b/config/drupal-vm.settings.php
@@ -32,3 +32,4 @@ $settings['jcms_sqs_queue'] = 'journal-cms--queue-local';
 $settings['jcms_sns_topic_template'] = 'arn:aws:sns:local:000000000000:%s';
 $settings['jcms_sqs_region'] = 'us-east-1';
 $settings['jcms_articles_endpoint'] = 'https://prod--gateway.elifesciences.org/articles/%s/versions';
+$settings['jcms_all_articles_endpoint'] = 'https://prod--gateway.elifesciences.org/articles';
--- a/src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
+++ b/src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
@@ -9,20 +9,59 @@
  * Implements hook_drush_command().
  */
 function jcms_notifications_drush_command() {
-  $items = array();
-  $items['article-import'] = array(
-    'description' => 'Imports articles from Lax.',
-    'drupal dependencies' => array('jcms_notifications'),
-    'aliases' => array('ai'),
-    'arguments' => array(
+  $items = [];
+  $items['article-import'] = [
+    'description' => 'Imports articles from Lax via SQS.',
+    'drupal dependencies' => ['jcms_notifications'],
+    'aliases' => ['ai'],
+    'arguments' => [
       'lrp' => 'Long running process or not. Defaults to false.',
-    ),
-    'examples' => array(
+    ],
+    'examples' => [
       'drush article-import 1' => 'Import articles from Lax as a long running process.',
       'drush article-import' => 'Import articles from Lax and return a message when finished.',
-    ),
-  );
+    ],
+  ];
+  $items['article-import-all'] = [
+    'description' => 'Imports all articles from Lax.',
+    'drupal dependencies' => ['jcms_notifications'],
+    'aliases' => ['aia'],
+    'examples' => [
+      'drush article-import-all' => 'Import articles from Lax and return a message when finished.',
+    ],
+  ];
   return $items;
+}
+
+/**
+ * Callback function drush_jcms_notifications_article_import_all().
+ */
+function drush_jcms_notifications_article_import_all() {
+  $fetch_service = \Drupal::service('jcms_article.fetch_article');
+  $fetch_versions_service = \Drupal::service('jcms_article.fetch_article_versions');
+  $crud_service = \Drupal::service('jcms_article.article_crud');
+  drush_print(dt('Fetching all article IDs. This may take a few minutes.'));
+  $ids = $fetch_service->getAllArticleIDs();
+  drush_print(dt('Received !count article IDs to process.', ['!count' => count($ids)]));
+  if ($ids) {
+    $time_start = microtime(TRUE);
+    foreach ($ids as $num => $id) {
+      $articleVersions = $fetch_versions_service->getArticleVersions($id);
+      $crud_service->crudArticle($articleVersions);
+      drush_print(dt('Processed article !article_id (!num of !count)', [
+        '!article_id' => $id,
+        '!num' => $num + 1,
+        '!count' => count($ids),
+      ]));
+    }
+    $time_end = microtime(TRUE);
+    $time = round($time_end - $time_start, 0);
+    drush_print(dt('Processed !count articles in !minutes minutes !seconds seconds.', [
+      '!count' => count($ids),
+      '!minutes' => floor($time / 60),
+      '!seconds' => round($time % 60),
+    ]));
+  }
 }
 
 /**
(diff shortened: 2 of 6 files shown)


src/modules/jcms_notifications/src/FetchArticleService.php
```diff
@@ -63,6 +63,32 @@ public function getArticle(string $id) {
...
+  public function getAllArticleIDs() {
+    $ids = [];
+    $endpoint = Settings::get('jcms_all_articles_endpoint');
+    if ($endpoint) {
+      $auth_key = 'jcms_article_auth_unpublished';
+      $response = $this->client->get($endpoint, ['auth' => Settings::get($auth_key), 'http_errors' => FALSE]);
+      if ($response instanceof ResponseInterface) {
+        $body = $response->getBody()->getContents();
+        $json = json_decode($body, TRUE);
+        if (isset($json['items'])) {
```
nlisgo: I believe this will only give us the articles on the first page (20):

https://github.com/elifesciences/api-raml/blob/develop/src/api.raml#L114

jamiehollern: Is there a way to get all articles on a single page and if there is, is that an okay way of doing things? Alternatively I can make it loop through the pages.

config/drupal-vm.settings.php
```diff
@@ -32,3 +32,4 @@
 $settings['jcms_sns_topic_template'] = 'arn:aws:sns:local:000000000000:%s';
 $settings['jcms_sqs_region'] = 'us-east-1';
 $settings['jcms_articles_endpoint'] = 'http://end2end--gateway.elifesciences.org/articles/{id}/versions';
+$settings['jcms_all_articles_endpoint'] = 'https://continuumtest--lax.elifesciences.org/api/v2/articles';
```
jamiehollern: @giorgiosironi This will need to be a value in your formula if you run this on the end2end tests.

jamiehollern:
Heads up to everyone, the Drush command to ingest all articles is `drush article-ingest` or `drush ain`.

src/modules/jcms_notifications/src/FetchArticleService.php
nlisgo: There is a way to get all articles on the same page. You can use a similar technique to `\Drupal\jcms_rest\Tests\Functional\RecursiveEndpointValidatorTest`: https://github.com/elifesciences/journal-cms/blob/develop/src/modules/jcms_rest/tests/src/Functional/RecursiveEndpointValidatorTest.php#L133-L143

nlisgo: If you need to introduce batch processing then it may just be as good to handle the results on a page in each batch.

jamiehollern:
@nlisgo Updated to run through all pages. Not batched as it'll be used as a Drush command but can be if we see timeouts.

src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
```diff
@@ -11,7 +11,7 @@
 function jcms_notifications_drush_command() {
   $items = array();
   $items['article-import'] = array(
```
giorgiosironi: `import` vs `ingest` does not differentiate between what the two tasks are doing, please rename to `article-import` (already correct) and `article-import-all`

config/drupal-vm.settings.php
giorgiosironi: Added to the formula, for both `end2end` and most importantly `prod`

jamiehollern:
@giorgiosironi Updated as per your request. Command is now `drush article-import-all` and shortcut is `drush aia`.

src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
```diff
@@ -22,10 +22,34 @@ function jcms_notifications_drush_command() {
...
+ */
+function drush_jcms_notifications_article_import_all() {
+  $fetch_service = \Drupal::service('jcms_notifications.fetch_article_service');
+  $crud_service = \Drupal::service('jcms_notifications.article_crud_service');
+  $ids = $fetch_service->getAllArticleIDs();
+  if ($ids) {
+    foreach ($ids as $id) {
+      $article = $fetch_service->getArticle($id);
+      $crud_service->crudArticle($article);
+      drush_print(dt('Processed article !article_id', ['!article_id' => $id]));
```
stephenwf: @jamiehollern Is there already a long running process for importing articles from SQS? On search @giorgiosironi implemented a change that basically just added all the articles to SQS and let the long running process deal with them. Is that something that would work in this case too?

src/modules/jcms_notifications/src/FetchArticleService.php
```diff
@@ -63,6 +63,43 @@ public function getArticle(string $id) {
...
+        $page = $page ?? 1;
+        $response = $this->client->get($endpoint, ['auth' => Settings::get($auth_key), 'http_errors' => FALSE, ['query' => ['page' => $page]]]);
+        if ($response instanceof ResponseInterface) {
+          $body = $response->getBody()->getContents();
+          $json = json_decode($body, TRUE);
+          if (isset($json['items']) && !empty($json['items'])) {
+            foreach ($json['items'] as $data) {
+              if (isset($data['id'])) {
+                $ids[] = $data['id'];
+                $page++;
```
stephenwf: Not quite sure I'm getting what this page variable is? Is it incrementing every single item, or just every set of items? It looks like it might be needed outside of the foreach

src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
jamiehollern: @stephenwf It probably would but we decided to grab them all from Lax so we didn't have to rely on SQS. What are your thoughts @giorgiosironi?

src/modules/jcms_notifications/src/FetchArticleService.php
jamiehollern: @stephenwf It's the pagination variable as we have to go through each page of results. I'll have a look and move it if necessary.

src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
giorgiosironi: Yes there is a process (`drush ai`), the problem we encountered in search was that the memory of the bulk import grows and grows. However, it may be the SDK that was filling it, while this project isn't using it (I think?). Testing this assumption: can you run this locally pointing to `prod--gateway.elifesciences.org/articles` and successfully import 3000+ articles without going out of memory?

stephenwf: I think the script will load all of the ids up front, and then request each individually, which should be okay for memory, not sure about how long it would take.

@jamiehollern Maybe we could use Guzzles promises, so something like (psdeudo):
```
$promises[] = $fetch_service->getArticle($id)->then(function($article) {
  $crud_service->crudArticle($article);
});
...

all($promises)->then(  ... done ... )
```

config/drupal-vm.settings.php
```diff
@@ -32,3 +32,4 @@
 $settings['jcms_sns_topic_template'] = 'arn:aws:sns:local:000000000000:%s';
 $settings['jcms_sqs_region'] = 'us-east-1';
 $settings['jcms_articles_endpoint'] = 'https://prod--gateway.elifesciences.org/articles/%s/versions';
+$settings['jcms_all_articles_endpoint'] = 'https://prod--gateway.elifesciences.org/articles';
```
nlisgo: Should we use a common prefix for both endpoints?

jamiehollern: Sure.

nlisgo:
This has been closed in favour of: https://github.com/elifesciences/journal-cms/pull/65

PR #51, opened 2016-12-21
ELPP-1758, ELPP-1839, ELPP-1840 Save article JSON when article is created and update published date and published status for articles.

jamiehollern:
@nlisgo @giorgiosironi 

Covers ELPP-1758, ELPP-1840 and ELPP-1839.

### DIFF
--- a/src/modules/jcms_notifications/src/ArticlePresave.php
+++ b/src/modules/jcms_notifications/src/ArticlePresave.php
@@ -1,0 +1,166 @@
+<?php
+
+namespace Drupal\jcms_notifications;
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\paragraphs\Entity\Paragraph;
+
+/**
+ * Class ArticlePresave.
+ *
+ * @package Drupal\jcms_notifications
+ */
+class ArticlePresave {
+
+  /**
+   * @var \Drupal\jcms_notifications\FetchArticleService
+   */
+  protected $fetchArticle;
+
+  /**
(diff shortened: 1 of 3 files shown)


src/modules/jcms_notifications/src/ArticlePresave.php
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * Sets the published date on the node.
+   */
+  public function setPublishedDate() {
+    if ($this->nodeIsValid()) {
+      $id = $this->getArticleIdFromNode();
+      $article = $this->getArticleById($id);
+      // Whether new or not, set the published date.
+      $json = json_decode($article['data']['unpublished']);
+      $date = date('U', $json['published']);
```
stephenwf: We've had some trouble with dates previously, might be worth checking if `$date` comes out as a `bool`

src/modules/jcms_notifications/src/ArticlePresave.php
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  /**
+   * Sets the published status of the node.
+   */
+  public function setPublishedStatus() {
+    if ($this->nodeIsValid()) {
+      $id = $this->getArticleIdFromNode();
+      $article = $this->getArticleById($id);
+      $json = json_decode($article['data']['unpublished']);
```
stephenwf: I don't think this is used

jamiehollern: Well spotted, sorted now.

PR #56, opened 2017-01-13
Fix for null image style objects.

jamiehollern:
@giorgiosironi @stephenwf

### DIFF
--- a/src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
+++ b/src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
@@ -145,7 +145,10 @@ abstract class AbstractRestResourceBase extends ResourceBase {
               str_replace(':', 'x', $ar),
               $width . 'x' . $height,
             ];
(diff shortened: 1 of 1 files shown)


giorgiosironi:
This didn't run because I locked `journal-cms--ci` for debugging, unblocking and building now.

giorgiosironi:
And it's now green on Jenkins, making the build stable in spite of an existing database. Other PRs should rebase or merge `develop` after this has got in.

jamiehollern:
@giorgiosironi Can we get this merged in then?

giorgiosironi:
I think so.

giorgiosironi:
By the way this coveralls bot does not seem to give us useful feedback, it's always red because coverage goes down 0.001%?

giorgiosironi:
(I don't see coverage as a particularly interesting metric for this project)

jamiehollern:
@giorgiosironi Agreed on both points. We can talk to @nlisgo about it when he's back perhaps.

stephenwf:
The way I've seen it being used on other projects is to tackle back logs of coverage with each merge, so even if the coverage for this commit doesn't need covered the whole project is raised. Might not fit our use-case though

PR #58, opened 2017-01-20
ELPP-1758 Improved the article import and CRUD.

jamiehollern:
@giorgiosironi @stephenwf This is the fixes for the issue with the refactor included. One thing to note for your formula is that I've changed the format of the `jcms_articles_endpoint` Drupal variable to be consistent (see [the diff](https://github.com/elifesciences/journal-cms/pull/58/files#diff-801dc01e1e6ce986ec8619e987d6ce95) for details).


config/drupal-vm.settings.php
```diff
@@ -31,4 +31,4 @@
 // Production template is 'arn:aws:sns:us-east-1:512686554592:bus-%s--dev'.
 $settings['jcms_sns_topic_template'] = 'arn:aws:sns:local:000000000000:%s';
 $settings['jcms_sqs_region'] = 'us-east-1';
-$settings['jcms_articles_endpoint'] = 'http://end2end--gateway.elifesciences.org/articles/{id}/versions';
+$settings['jcms_articles_endpoint'] = 'https://prod--gateway.elifesciences.org/articles/%s/versions';
```
giorgiosironi: ok, I will update it in the formula right before merge from `{id}` to `%s`.

src/modules/jcms_article/src/ArticleCrud.php
```diff
@@ -0,0 +1,145 @@
...
+    $node = NULL;
+    if ($nid = $this->getNodeIdByArticleId($articleVersions->getId())) {
+      if ($articleVersions->getAction() == ArticleVersions::DELETE) {
+        $node = $this->deleteArticle($nid);
+      }
+      else {
+        $node = $this->updateArticle($articleVersions);
+      }
+    }
+    return $node;
```
giorgiosironi: the docblock says this method should also create the article node, but there is no call to `createArticle()`?

src/modules/jcms_article/src/ArticleCrud.php
```diff
@@ -0,0 +1,145 @@
...
+   * Helper method to decide whether to created, update or delete a node.
+   *
+   * @param \Drupal\jcms_article\Entity\ArticleVersions $articleVersions
+   *
+   * @return \Drupal\Core\Entity\EntityInterface|null
+   */
+  public function crudArticle(ArticleVersions $articleVersions) {
+    $node = NULL;
+    if ($nid = $this->getNodeIdByArticleId($articleVersions->getId())) {
+      if ($articleVersions->getAction() == ArticleVersions::DELETE) {
```
giorgiosironi: are we ever going to delete articles? Maybe we should remove that part

src/modules/jcms_article/src/ArticleCrud.php
jamiehollern: Well spotted, it must have been removed when I rebased develop into it.

src/modules/jcms_article/src/ArticleCrud.php
jamiehollern: According to @nlisgo if we try to access an article and it returns a 404 then it should be deleted. It may never happen but it may be useful to retain?

src/modules/jcms_article/src/Entity/ArticleVersions.php
```diff
@@ -0,0 +1,140 @@
...
+ * Class Article
+ *
+ * @package Drupal\jcms_article\Entity
+ */
+final class ArticleVersions {
+
+  const WRITE = 1;
+
+  const DELETE = 2;
+
```
giorgiosironi: @lsh-0 please confirm these two are the only possible values for `stage` in article versions

lsh-0: um - 1 or 2? no. `preview` or `published`: https://github.com/elifesciences/api-raml/blob/develop/dist/model/article-poa.v1.json#L62

src/modules/jcms_article/src/FetchArticleVersions.php
```diff
@@ -0,0 +1,86 @@
...
+   * Makes the request to get the article versions.
+   *
+   * @param string $id
+   *
+   * @return \Psr\Http\Message\ResponseInterface
+   * @throws \TypeError
+   */
+  function requestArticleVersions(string $id): ResponseInterface {
+    $options = [
+      'auth' => Settings::get('jcms_article_auth_unpublished'),
```
giorgiosironi: ok, always use this authentication

src/modules/jcms_article/src/Entity/ArticleVersions.php
giorgiosironi: yes, look at the constants `PUBLISHED` and `UNPUBLISHED`

src/modules/jcms_article/src/Entity/ArticleVersions.php
```diff
@@ -0,0 +1,140 @@
...
+  const PUBLISHED = 'published';
+
+  const UNPUBLISHED = 'preview';
```
jamiehollern: @lsh-0 These are the constants @giorgiosironi is talking about.

lsh-0: yeah, they look about right.

giorgiosironi:
I commented on what I know, but at this point we should also wait for the review of @nlisgo on Monday so that we are all on the same page.

jamiehollern:
@giorgiosironi Are we waiting for Nathan to get back before merging this? It would be ideal to get it in now to try and clear some stuff and unblock some other tickets.

src/modules/jcms_article/tests/src/Unit/Entity/test.json
```diff
@@ -0,0 +1,108 @@
...
+      },
+      "impactStatement": "RID is a descending peptidergic neuron that sustains the forward motor state.",
+      "title": "Neuroendocrine modulation sustains the <i>C. elegans</i> forward motor state",
+      "authorLine": "Maria A Lim et al",
+      "volume": 5
+    }
+  ],
+  "accepted": "2016-11-14",
+  "received": "2016-07-21"
+}
```
nlisgo: Nitpick: newline at end of file

src/modules/jcms_article/src/Hooks/NodePresave.php
```diff
@@ -0,0 +1,180 @@
...
+   */
+  public function setPublishedDate(EntityInterface $entity) {
+    $id = $entity->label();
+    $article = $this->getArticleById($id);
+    // Use the published JSON if no unpublished exists.
+    $version = $article->getLatestUnpublishedVersionJson() ?: $article->getLatestPublishedVersionJson();
+    $json = json_decode($version);
+    if (property_exists($json, 'versionDate')) {
+      $date = strtotime($json->versionDate);
+      $entity->set('created', $date);
```
nlisgo: The created date should be set to the published date of an article which is the same for all versions. This will meet the requirement in ELPP-1839.

There is an additional requirement that can be handled in a separate jira ticket to store the statusDate of the latest published version of an article which can be used for sorting appropriately.

giorgiosironi: Agreed - as far as I know this date is taken from the original JATS XML and should always be present even in an article with a single version in `preview` (/cc @lsh-0)

giorgiosironi:
Rerunning build now that api-dummy should be in place.

src/modules/jcms_article/src/Hooks/NodePresave.php
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  /**
+   * Sets the published status of the node.
+   */
+  public function setPublishedStatus(EntityInterface $entity) {
+    $id = $entity->label();
+    $article = $this->getArticleById($id);
+    // Use the published JSON if no unpublished exists.
+    $version = $article->getLatestUnpublishedVersionJson() ?: $article->getLatestPublishedVersionJson();
```
nlisgo: The published status of the article node should be set to 1 if there are any published versions of an article and set to 0 if there are not published versions of an article.

This would set the article to unpublished if the latest version of an article is in preview. This is not the expected behaviour.

src/modules/jcms_article/src/Hooks/NodePresave.php
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  /**
+   * Sets the article subjects on the article as taxonomy terms.
+   */
+  public function setSubjectTerms(EntityInterface $entity) {
+    $id = $entity->label();
+    $article = $this->getArticleById($id);
+    // Use the published JSON if no unpublished exists.
+    $version = $article->getLatestUnpublishedVersionJson() ?: $article->getLatestPublishedVersionJson();
```
nlisgo: All values for the article node should reflect the latest published version of an article, if available. Use the latest preview article version if no published are available.

src/modules/jcms_article/src/Hooks/NodePresave.php
giorgiosironi: I agree with the expected behavior

src/modules/jcms_article/src/Hooks/NodePresave.php
nlisgo: @giorgiosironi it is available in the JATS XML but I believe that it is stripped out in the article json.

See: https://github.com/elifesciences/api-raml/blob/develop/src/snippets/article.v1.yaml#L49

giorgiosironi: ah, didn't know that, thanks

PR #67, opened 2017-01-30
Performing smoke tests on all APIs on deploy

giorgiosironi:


giorgiosironi:
I get an error locally, and we are having some similar problem on `/covers` in production. Let's see if we can make this green.

nlisgo:
I will look at this tomorrow.

giorgiosironi:
This has become green on `ci` without changing the code, so it may depend on the state of the db. Which is a good reason to merge it so that we can verify it on all envs.

Unrolling the loop now.

smoke_tests.sh
```diff
@@ -1,4 +1,19 @@
...
+hostname=${1:-$local_hostname}
+
+echo "Homepage"
+[ $(curl --write-out %{http_code} --silent --output /dev/null "$hostname") == 200 ]
+
+echo "APIs"
+[ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.blog-article-list+json; version=1" "${hostname}/blog-articles") == 200 ]
+[ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.collection-list+json; version=1" "${hostname}/collections") == 200 ]
+[ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.cover-list+json; version=1" "${hostname}/covers") == 200 ]
+[ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.event-list+json; version=1" "$hostname"/"${hostname}/events") == 200 ]
```
nlisgo: You have repeated `$hostname`:

```
[ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.event-list+json; version=1" "$hostname"/"${hostname}/events") == 200 ]
```

giorgiosironi: sorry, too many amendments, correcting it now

PR #68, opened 2017-01-30
Switch to using drupal packagist and update drupal-vm

nlisgo:
@jamiehollern 

See: https://packagist.drupal-composer.org/


nlisgo:
I successfully ran the following locally:

```
vagrant destroy
composer clean-up
COMPOSER=composer-setup.json composer install
vagrant up
```

PR #71, opened 2017-02-01
Save recent covers to entityqueue and other improvements

nlisgo:
@jamiehollern

### DIFF
--- a/src/modules/jcms_migrate/src/EventSubscriber/JCMSMIgrateEventSubscriber.php
+++ b/src/modules/jcms_migrate/src/EventSubscriber/JCMSMIgrateEventSubscriber.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Drupal\jcms_migrate\EventSubscriber;
+
+use Drupal\entityqueue\Entity\EntitySubqueue;
+use Drupal\migrate\Event\MigrateEvents;
+use Drupal\migrate\Event\MigratePostRowSaveEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class JCMSMIgrateEventSubscriber implements EventSubscriberInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
+    $events[MigrateEvents::POST_ROW_SAVE] = 'onPostRowSave';
+    return $events;
+  }
+
+  /**
+   * Code to run after a row has been saved.
+   */
+  public function onPostRowSave(MigratePostRowSaveEvent $event) {
+    if ($event->getMigration()->id() == 'jcms_covers_db') {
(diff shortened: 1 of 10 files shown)


src/modules/jcms_migrate/src/EventSubscriber/JCMSMIgrateEventSubscriber.php
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  /**
+   * Code to run after a row has been saved.
+   */
+  public function onPostRowSave(MigratePostRowSaveEvent $event) {
+    if ($event->getMigration()->id() == 'jcms_covers_db') {
+      if ($event->getRow()->getDestinationProperty('counter') <= 3) {
+        $subqueue = EntitySubqueue::load('covers');
+        $items = $subqueue->get('items')->getValue();
```
jamiehollern: For these sorts of things, I tend to be more defensive. Unless you're sure `getValue()` won't run on `null` I'd maybe separate this. It's not a huge issue though, if you're confident in it.

src/modules/jcms_rest/src/Plugin/rest/resource/PersonItemRestResource.php
```diff
@@ -71,23 +71,39 @@ public function get($id) {
...
+          $expertises = [
+            'id' => [],
+            'name' => [],
+          ];
           foreach ($research_details_field->get('field_research_expertises') as $expertise) {
-            $research['expertises'][] = [
-              'id' => $expertise->get('entity')->getValue()->get('field_subject_id')->getString(),
-              'name' => $expertise->get('entity')->getValue()->toLink()->getText(),
-            ];
+            $expertise_id = $expertise->get('entity')->getValue()->get('field_subject_id')->getString();
```
jamiehollern: Same as the above comment, this type of method chaining can be dangerous unless you're confident in it.

src/modules/jcms_migrate/src/EventSubscriber/JCMSMIgrateEventSubscriber.php
nlisgo: Will make this adjustment. This one will certainly cause a problem.

src/modules/jcms_rest/src/Plugin/rest/resource/PersonItemRestResource.php
nlisgo: Will adjust

nlisgo: Actually might refactor this later. This doesn't make the code any worse than it was and if we are going to introduce these checks we should probably be consistent. We could do this as a clean up step covered by a separate jira ticket.

PR #75, opened 2017-02-02
Allow cover list to be filtered by date range

nlisgo:


src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
```diff
@@ -322,6 +324,22 @@ protected function filterPageAndOrder(QueryInterface &$query, $sort_by = NULL) {
...
+   * @param \Drupal\Core\Entity\Query\QueryInterface $query
+   */
+  protected function filterDateRange(QueryInterface &$query) {
+    $start_date = $this->getRequestOption('start-date');
+    if (!empty($start_date)) {
+      $query->condition('created', strtotime($start_date), '>=');
+    }
+    $end_date = $this->getRequestOption('end-date');
+    if (!empty($start_date)) {
+      $query->condition('created', strtotime($end_date), '<');
```
thewilkybarkid: `<=`

thewilkybarkid: `<=`

src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
```diff
@@ -322,6 +324,22 @@ protected function filterPageAndOrder(QueryInterface &$query, $sort_by = NULL) {
...
 
   /**
+   * Apply filter for date range by amending query.
+   *
+   * @param \Drupal\Core\Entity\Query\QueryInterface $query
+   */
+  protected function filterDateRange(QueryInterface &$query) {
+    $start_date = $this->getRequestOption('start-date');
+    if (!empty($start_date)) {
+      $query->condition('created', strtotime($start_date), '>=');
```
thewilkybarkid: My validation/date creation comment in https://github.com/elifesciences/search/pull/103 apply here too.

nlisgo:
@thewilkybarkid good spot on the filter. I'd fixed it in one place but not the other. Now sharing the code :)

nlisgo:
@thewilkybarkid is the documentation on what the Content-Type header should be for a Bad Request response?

thewilkybarkid:
All errors need to be `application/problem+json` (like https://github.com/elifesciences/api-raml/blob/develop/src/api.raml#L77-L82).

PR #79, opened 2017-02-06
ELPP-1815 Improved error handling in long running process.

jamiehollern:
@giorgiosironi @nlisgo

### DIFF
--- a/src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
+++ b/src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
@@ -5,6 +5,8 @@
  * Drush commands for the module.
  */
 
+use GuzzleHttp\Exception\RequestException;
+
 /**
  * Implements hook_drush_command().
  */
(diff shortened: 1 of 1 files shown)


src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
```diff
@@ -95,12 +97,24 @@ function drush_jcms_notifications_article_import($lrp = FALSE) {
...
-        drush_print(dt('Deleted from the queue message !message_id', ['!message_id' => $message->getMessageId()]));
-        $count++;
+        try {
+          $articleVersions = $fetch_service->getArticleVersions($id);
+          $crud_service->crudArticle($articleVersions);
+          drush_print(dt('Processed message !message_id', ['!message_id' => $message->getMessageId()]));
+        }
+        catch (Exception $e) {
+          if ($e instanceof RequestException) {
+            error_log($e);
```
giorgiosironi: `error_log` usually accepts strings. Can you extract a few things like `$e->getMessage()`, `$e->getTraceAsString()`

giorgiosironi: I would also log all `$e`, not just the `RequestException`.

src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
```diff
@@ -95,12 +97,24 @@ function drush_jcms_notifications_article_import($lrp = FALSE) {
...
+          }
+          else {
+            throw $e;
+          }
+        }
+        finally {
+          $queue_service->deleteMessage($message);
```
giorgiosironi: the `finally` ensures that there can be no continuous retry of the same message, nice

src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
giorgiosironi: or on the contrary:
- `RequestException` with a 404 are unpublished content, can be ignored
- `RequestException` in general are logged
- all other exceptions are logged

jamiehollern:
@giorgiosironi Updated the code. All exceptions will be logged because thanks to our other work there should be no exceptions caused by 404s/500s etc.

PR #80, opened 2017-02-08
ELPP-2107 Removed dblog module from module configuration.

jamiehollern:
@giorgiosironi As discussed.

@nlisgo Below is a list of additional modules I think we should disable. Some are used in development but they should be disabled in production so we can enable them at the time they're needed. The performance gains would be marginal but it also keeps the install clean and small.

- automated_cron
- color
- comment
- contact
- field_ui
- hal
- history
- rdf
- search
- tour
- views_ui

### DIFF
--- a/sync/core.extension.yml
+++ b/sync/core.extension.yml
@@ -17,7 +17,6 @@ module:
   ctools: 0
   datetime: 0
   datetime_range: 0
-  dblog: 0
   devel_generate: 0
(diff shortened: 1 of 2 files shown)


giorgiosironi:
I like the removal, let's wait for @nlisgo to see if we can remove something else too.

nlisgo:
I think you should reexport the configuration as part of this pull request.

PR #81, opened 2017-02-08
feature/ELPP- 1815

jamiehollern:
@giorgiosironi


src/modules/jcms_article/src/ArticleCrud.php
```diff
@@ -182,7 +182,11 @@ public function createParagraph(EntityInterface $node, ArticleVersions $articleV
    */
   public function deleteArticle(ArticleVersions $articleVersions) {
```
giorgiosironi: let's keep open a discussion on deleteArticle() btw; we probably should never delete this type of content as Lax won't

giorgiosironi:
Merging when the builds are finished.

PR #82, opened 2017-02-08
feature/ELPP-1456 Bugfixes.

jamiehollern:
@giorgiosironi Fixed 3 bugs.

### DIFF
--- a/src/modules/jcms_notifications/src/MysqlNotificationStorage.php
+++ b/src/modules/jcms_notifications/src/MysqlNotificationStorage.php
@@ -3,6 +3,7 @@
 namespace Drupal\jcms_notifications;
 
 use Drupal\Core\Database\Driver\mysql\Connection;
+use Drupal\Core\Entity\EntityInterface;
 
 /**
  * Class MysqlNotificationStorage.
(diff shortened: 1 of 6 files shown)


giorgiosironi:
Had some problem in running on Jenkins, now green. Given the size I'd leave this to @nlisgo, outside of my knowledge of Drupal.

src/modules/jcms_notifications/src/MysqlNotificationStorage.php
```diff
@@ -35,8 +36,13 @@ public function __construct(Connection $connection) {
   /**
    * @inheritdoc
    */
-  public function saveNotificationNid(int $nodeId) {
-    return $this->connection->insert(self::TABLE)->fields(['node_id'], [$nodeId])->execute();
+  public function saveNotificationNid(EntityInterface $entity) {
```
nlisgo: Should we not rename the method as we are not passing nid anymore?

jamiehollern: No, as it still saves the `nid`. The name is based on what it does rather than what it receives.

PR #83, opened 2017-02-09
Optimizing external calls on article-import-all

giorgiosironi:
Before the change, the static cache was not working very well as each article being imported was causing 5 requests to the /articles/:id/version API. After the change, the number of calls is reduced to 2, one in the service and only one from the NodePresave object.

Moreover, passing the article object on the stack is safer than keeping it around as a global static field, which may distribute stale data in long-running processes

### DIFF
--- a/src/modules/jcms_article/jcms_article.module
+++ b/src/modules/jcms_article/jcms_article.module
@@ -32,8 +32,10 @@ function jcms_article_node_presave(EntityInterface $entity) {
     return NULL;
   }
   $node_presave = \Drupal::service('jcms_article.hooks.node_presave');
-  $node_presave->addJsonFields($entity);
-  $node_presave->setPublishedStatus($entity);
-  $node_presave->setStatusDate($entity);
-  $node_presave->setSubjectTerms($entity);
+  // TODO: retrieve with getArticleById() and pass it in
+  $article = $node_presave->getArticleById($entity->label());
+  $node_presave->addJsonFields($entity, $article);
+  $node_presave->setPublishedStatus($entity, $article);
+  $node_presave->setStatusDate($entity, $article);
+  $node_presave->setSubjectTerms($entity, $article);
 }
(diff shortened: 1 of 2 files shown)


src/modules/jcms_article/jcms_article.module
```diff
@@ -32,8 +32,10 @@ function jcms_article_node_presave(EntityInterface $entity) {
...
-  $node_presave->addJsonFields($entity);
-  $node_presave->setPublishedStatus($entity);
-  $node_presave->setStatusDate($entity);
-  $node_presave->setSubjectTerms($entity);
+  // TODO: retrieve with getArticleById() and pass it in
+  $article = $node_presave->getArticleById($entity->label());
+  $node_presave->addJsonFields($entity, $article);
+  $node_presave->setPublishedStatus($entity, $article);
+  $node_presave->setStatusDate($entity, $article);
+  $node_presave->setSubjectTerms($entity, $article);
```
nlisgo: It seems wrong to have to keep on adding the same `$article` value. Get how we should avoid multiple requests but surely this could be handled by ensuring that we retained the results of previews requests when calling `getArticleById()`.

Perhaps we should discuss.

giorgiosironi: Since we don't know for how many hours/days the NodePresave objects has been around, anything cached inside the object or static can in principle be stale. Moreover, storing responses bloats memory (I get daily alerts `Memory > 90% on 'journal-cms--prod--1` at the moment). An ideal way to avoid the multiple parameter passing would be to create a new NodePresave object for each call to this method.

nlisgo:
Needs a rebase

jamiehollern:
@nlisgo I don't have write access to this repo so I can't apply the rebase to this PR. Therefore I've opened https://github.com/elifesciences/journal-cms/pull/86.

nlisgo:
Rebase has been done here: https://github.com/elifesciences/journal-cms/pull/86

PR #86, opened 2017-02-15
Optimizing external calls on article-import-all

jamiehollern:
Before the change, the static cache was not working very well as each article being imported was causing 5 requests to the /articles/:id/version API. After the change, the number of calls is reduced to 2, one in the service and only one from the NodePresave object.

Moreover, passing the article object on the stack is safer than keeping it around as a global static field, which may distribute stale data in long-running processes

### DIFF
--- a/src/modules/jcms_article/jcms_article.module
+++ b/src/modules/jcms_article/jcms_article.module
@@ -32,8 +32,9 @@ function jcms_article_node_presave(EntityInterface $entity) {
     return NULL;
   }
   $node_presave = \Drupal::service('jcms_article.hooks.node_presave');
-  $node_presave->addJsonFields($entity);
-  $node_presave->setPublishedStatus($entity);
-  $node_presave->setStatusDate($entity);
-  $node_presave->setSubjectTerms($entity);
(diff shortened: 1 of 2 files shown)


giorgiosironi:
Thank you for the rebase (I am working on other stuff), I know @nlisgo expressed come concern at this pattern in the other PR.

src/modules/jcms_article/jcms_article.module
```diff
@@ -32,8 +32,10 @@ function jcms_article_node_presave(EntityInterface $entity) {
     return NULL;
   }
   $node_presave = \Drupal::service('jcms_article.hooks.node_presave');
-  $node_presave->addJsonFields($entity);
-  $node_presave->setPublishedStatus($entity);
-  $node_presave->setStatusDate($entity);
-  $node_presave->setSubjectTerms($entity);
+  // TODO: retrieve with getArticleById() and pass it in
```
nlisgo: Can this todo be removed?

jamiehollern: @nlisgo Done.

PR #88, opened 2017-02-15
ELPP-1459 Added article fragment image support.

jamiehollern:
@nlisgo @giorgiosironi

### DIFF
--- a/config/drupal-vm.settings.php
+++ b/config/drupal-vm.settings.php
@@ -31,6 +31,8 @@ $settings['jcms_sqs_queue'] = 'journal-cms--queue-local';
 // Production template is 'arn:aws:sns:us-east-1:512686554592:bus-%s--dev'.
 $settings['jcms_sns_topic_template'] = 'arn:aws:sns:local:000000000000:%s';
 $settings['jcms_sqs_region'] = 'us-east-1';
-$settings['jcms_articles_endpoint'] = 'https://prod--gateway.elifesciences.org/articles/%s/versions';
-$settings['jcms_all_articles_endpoint'] = 'https://prod--gateway.elifesciences.org/articles';
+$settings['jcms_gateway'] = 'https://prod--gateway.elifesciences.org';
+$settings['jcms_all_articles_endpoint'] = $settings['jcms_gateway'] . '/articles';
+$settings['jcms_articles_endpoint'] = $settings['jcms_gateway'] . '/articles/%s/versions';
+$settings['jcms_article_fragment_images_endpoint'] = $settings['jcms_gateway'] . '/articles/%s/fragments/image';
 $settings['jcms_article_auth_unpublished'] = NULL;
--- a/src/modules/jcms_article_fragment/src/ArticleFragmentAccessControlHandler.php
+++ b/src/modules/jcms_article_fragment/src/ArticleFragmentAccessControlHandler.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Drupal\jcms_article_fragment;
+
+use Drupal\Core\Entity\EntityAccessControlHandler;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Access\AccessResult;
+
+/**
+ * Access controller for the Article fragment entity.
+ *
+ * @see \Drupal\jcms_article_fragment\Entity\ArticleFragment.
+ */
+class ArticleFragmentAccessControlHandler extends EntityAccessControlHandler {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function checkAccess(EntityInterface $entity, $operation, AccountInterface $account) {
+    /** @var \Drupal\jcms_article_fragment\Entity\ArticleFragmentInterface $entity */
+    switch ($operation) {
+      case 'view':
+        if (!$entity->isPublished()) {
+          return AccessResult::allowedIfHasPermission($account, 'view unpublished article fragment entities');
+        }
+        return AccessResult::allowedIfHasPermission($account, 'view published article fragment entities');
+
+      case 'update':
+        return AccessResult::allowedIfHasPermission($account, 'edit article fragment entities');
+
+      //case 'delete':
+      //  return AccessResult::allowedIfHasPermission($account, 'delete article fragment entities');
+    }
+
+    // Unknown operation, no opinion.
+    return AccessResult::neutral();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function checkCreateAccess(AccountInterface $account, array $context, $entity_bundle = NULL) {
+    return AccessResult::allowedIfHasPermission($account, 'add article fragment entities');
+  }
+
+}
--- a/src/modules/jcms_article_fragment/src/ArticleFragmentStorageInterface.php
+++ b/src/modules/jcms_article_fragment/src/ArticleFragmentStorageInterface.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Drupal\jcms_article_fragment;
+
+use Drupal\Core\Entity\ContentEntityStorageInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Language\LanguageInterface;
+use Drupal\jcms_article_fragment\Entity\ArticleFragmentInterface;
+
+/**
+ * Defines the storage handler class for Article fragment entities.
+ *
+ * This extends the base storage class, adding required special handling for
+ * Article fragment entities.
+ *
+ * @ingroup jcms_article_fragment
+ */
+interface ArticleFragmentStorageInterface extends ContentEntityStorageInterface {
+
+  /**
+   * Gets a list of Article fragment revision IDs for a specific Article fragment.
+   *
+   * @param \Drupal\jcms_article_fragment\Entity\ArticleFragmentInterface $entity
+   *   The Article fragment entity.
+   *
+   * @return int[]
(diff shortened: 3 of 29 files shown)


config/drupal-vm.settings.php
```diff
@@ -31,6 +31,8 @@
 // Production template is 'arn:aws:sns:us-east-1:512686554592:bus-%s--dev'.
 $settings['jcms_sns_topic_template'] = 'arn:aws:sns:local:000000000000:%s';
 $settings['jcms_sqs_region'] = 'us-east-1';
-$settings['jcms_articles_endpoint'] = 'https://prod--gateway.elifesciences.org/articles/%s/versions';
-$settings['jcms_all_articles_endpoint'] = 'https://prod--gateway.elifesciences.org/articles';
+$settings['jcms_gateway'] = 'https://prod--gateway.elifesciences.org';
+$settings['jcms_all_articles_endpoint'] = $settings['jcms_gateway'] . '/articles';
+$settings['jcms_articles_endpoint'] = $settings['jcms_gateway'] . '/articles/%s/versions';
+$settings['jcms_article_fragment_images_endpoint'] = $settings['jcms_gateway'] . '/articles/%s/fragment/image';
```
giorgiosironi: we need to add `jcms_article_fragment_images_endpoint` in the settings provided by the formula, also `jcms_gateway`?

giorgiosironi: `fragment` has been renamed to `fragments` for consistency with `/articles` and `/versions`

src/modules/jcms_article_fragment/src/ArticleFragmentAccessControlHandler.php
```diff
@@ -0,0 +1,47 @@
...
+      case 'view':
+        if (!$entity->isPublished()) {
+          return AccessResult::allowedIfHasPermission($account, 'view unpublished article fragment entities');
+        }
+        return AccessResult::allowedIfHasPermission($account, 'view published article fragment entities');
+
+      case 'update':
+        return AccessResult::allowedIfHasPermission($account, 'edit article fragment entities');
+
+      case 'delete':
```
giorgiosironi: The API does not support the deletion of fragments, hence the user probably can't delete them

src/modules/jcms_article_fragment/src/ArticleFragmentStorageInterface.php
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The user entity.
+   *
+   * @return int[]
+   *   Article fragment revision IDs (in ascending order).
+   */
+  public function userRevisionIds(AccountInterface $account);
+
+  /**
```
giorgiosironi: I wonder how i18n got into the requirements... is it just a standard interface?

src/modules/jcms_article_fragment/src/Form/ArticleFragmentForm.php
```diff
@@ -0,0 +1,109 @@
...
+    if (!$form_state->isValueEmpty('new_revision') && $form_state->getValue('new_revision') != FALSE) {
+      $entity->setNewRevision();
+      // If a new revision is created, save the current user as revision author.
+      $entity->setRevisionCreationTime(REQUEST_TIME);
+      $entity->setRevisionUserId(\Drupal::currentUser()->id());
+    }
+    else {
+      $entity->setNewRevision(FALSE);
+    }
+    $status = parent::save($form, $form_state);
```
giorgiosironi: ok, so this first saves the data locally, and then POSTs to Lax. Any error in the former stops the latter from happening.

src/modules/jcms_article_fragment/src/Form/ArticleFragmentForm.php
```diff
@@ -0,0 +1,109 @@
...
+   *
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   */
+  public function setImageFragment(FormStateInterface $form_state) {
+    $values = $form_state->getValues();
+    $article_id = $values['name'][0]['value'] ?? '';
+    $fid = $values['image'][0]['fids'][0] ?? 0;
+    if ($fid && $article_id) {
+      $this->api->postImageFragment($fid, $article_id);
+    }
```
giorgiosironi: else? exception?

giorgiosironi: `postImageFragment()` returns a response; it must be checked so that if it's not 200 an error should be shown. Anything different from 200 means the fragment is in this intermediate state in which is stored in journal-cms, but has not been sent to Lax and hence it won't appear on the article page. There should be a way to distinguish between a fragment that has been successfully set on Lax and one that isn't.

Given we are working with remote procedure calls and Lax is the owner of the article data, maybe we should switch around the order: first store in Lax, only if successful there then store it on journal-cms; any error in Lax interrupts the whole flow.

src/modules/jcms_article_fragment/src/Form/ArticleFragmentForm.php
giorgiosironi: see postImageFragment() for another possibility

src/modules/jcms_article_fragment/src/ArticleFragmentStorageInterface.php
jamiehollern: This is just a standard Drupal thing.

src/modules/jcms_article_fragment/src/ArticleFragmentAccessControlHandler.php
jamiehollern: This is just for the UI, is it okay to leave this here?

giorgiosironi: The problem is, if the underlying API does not allow to delete fragments (only to overwrite them), how can the user actually see a Delete button in the UI?

jamiehollern: It deletes the entity from Drupal.

src/modules/jcms_article_fragment/src/Form/ArticleFragmentForm.php
```diff
@@ -0,0 +1,117 @@
...
+    $entity = $this->entity;
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function save(array $form, FormStateInterface $form_state) {
+    // Post image fragment to Lax.
+    try {
```
jamiehollern: @giorgiosironi Guzzle will now throw exceptions to any non 2xx response, so we catch them here and set a message.

src/modules/jcms_article_fragment/src/ArticleFragmentAccessControlHandler.php
giorgiosironi: Yes, but what happens next is:
- the entity is deleted from Drupal
- Lax is not called (unless we patch it)
- the article page on elifesciences.org continues to display the image that has been "deleted"
- the user doesn't see it in the administration panel, getting confused

src/modules/jcms_article_fragment/src/Form/ArticleFragmentForm.php
giorgiosironi: nice, please log the exception message and stack trace somewhere (I guess `error_log` unless there is a more specialized logging service)

src/modules/jcms_article_fragment/src/ArticleFragmentAccessControlHandler.php
nlisgo: It may be that we need to adjust the lax API so that we can remove images. So, for now we can reject an attempt to delete an image. And complete this work when deletion of image is supported on lax.

nlisgo:
Some possible improvements we would want to make that could be handled in a separate piece of work:

- We ought to be able to upload a separate image for the listing view or the hero image.
- If we have uploaded article images to lax independent of journal-cms then when we pull down the article snippet we should also populate the image into Drupal.
- The ability to delete an article listing image.
- The ability to delete an article hero image.
- Lax should support storage of focal point coordinates for both listing and hero images.

src/modules/jcms_article_fragment/src/Form/ArticleFragmentForm.php
jamiehollern: I've updated this based on your other suggestion.

config/drupal-vm.settings.php
jamiehollern: @giorgiosironi Yes, please add those to the formula.

src/modules/jcms_article_fragment/src/ArticleFragmentAccessControlHandler.php
jamiehollern: I've commented out the code that allows deletion so that we can reimplement it later.

src/modules/jcms_article_fragment/src/Form/ArticleFragmentForm.php
jamiehollern: Implemented the above solution.

src/modules/jcms_article_fragment/src/Form/ArticleFragmentForm.php
jamiehollern: Done.

giorgiosironi:
Nothing stops this from being merged right?

jamiehollern:
@giorgiosironi It's ready to be merged pending your approval.

PR #89, opened 2017-02-15
Refactor highlights

nlisgo:


src/modules/jcms_rest/src/Plugin/rest/resource/HighlightListRestResource.php
```diff
@@ -0,0 +1,119 @@
...
+          $response = [];
+          foreach ($items as $item) {
+            $response[] = $this->getHighlightItem($item);
+          }
+        }
+      }
+    }
+
+    if (!empty($response)) {
+      $response = new JsonResponse($response, Response::HTTP_OK, ['Content-Type' => 'application/vnd.elife.highlights+json;version=1']);
```
jamiehollern: Not for this PR, but we should probably have some sort of static variable/constant for when referencing these content types.

giorgiosironi: If you use require `elife/api-client` through Composer you will have constants for all those (and also PHP classes to do the requests)

src/modules/jcms_rest/src/Plugin/rest/resource/HighlightListRestResource.php
```diff
@@ -0,0 +1,121 @@
...
+      return $response;
+    }
+
+    throw new JCMSNotFoundHttpException(t('Highlights list with ID @id was not found', ['@id' => $list]));
```
giorgiosironi: The api-raml seems to allow for an empty list:
https://github.com/elifesciences/api-raml/blob/develop/src/model/highlights.v1.yaml#L3
So shouldn't this case be a 200 response with an empty array in the `items` field?

nlisgo:
For now I am returning an empty array if no results are found.

We possibly want to return a 404 if the list id is not recognised or it might be ok to just return an empty array in all cases.

PR #90, opened 2017-02-16
Add published date to collections

nlisgo:


jamiehollern:
Seems this causes the tests to fail.

PR #92, opened 2017-02-16
Log type and id of content items just notified

giorgiosironi:
@jamiehollern how do we make it log the actual id sent in the notifications (e.g. `ba88c0df`)?

### DIFF
--- a/src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
+++ b/src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
@@ -159,8 +159,8 @@ function drush_jcms_notifications_send_notifications() {
     $nids = [];
     foreach ($nodes as $node) {
       $nids[] = $node->id();
-      $sns_crud->sendMessage($node);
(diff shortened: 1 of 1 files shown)


jamiehollern:
@giorgiosironi We can use the `getMessageFromNode` method to get the message object with the ID.

```
$message = $sns_crud->getMessageFromNode($node);
$id = $message->getId();
```

src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
```diff
@@ -160,7 +160,7 @@ function drush_jcms_notifications_send_notifications() {
     foreach ($nodes as $node) {
       $nids[] = $node->id();
       $sns_crud->sendMessage($node);
```
jamiehollern: The `sendMessage` method returns a `BusOutgoingMessage` object. You can get the ID from this.
```
$message = $sns_crud->sendMessage($node);
$id = $message->getId();
```

giorgiosironi: even easier, updating

PR #94, opened 2017-02-20
Check that goaws is not installed before installing

nlisgo:


nlisgo:
@jamiehollern I'll hold off merging this in until you can take a look.

PR #96, opened 2017-02-23
Added Redis smoke test.

jamiehollern:
@nlisgo @giorgiosironi


smoke_tests.sh
```diff
@@ -17,3 +17,6 @@ echo "APIs"
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.person-list+json; version=1" "${hostname}/people") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.podcast-episode-list+json; version=1" "${hostname}/podcast-episodes") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.subject-list+json; version=1" "${hostname}/subjects") == 200 ]
+
+echo "Redis"
+[ -z $(php -r '$redis = new \Redis();$redis->connect("$argv[1]", 6379);' "$hostname") ]
```
giorgiosironi: `php` will exit with a status code that is different from 0 in case of exception, so the whole line could just be:
```
 php -r '$redis = new \Redis(); $redis->connect($argv[1], 6379);' "$hostname"
```

giorgiosironi: nice by the way, because it tests both the extension and the daemon

jamiehollern: Good shout. Have updated.

PR #101, opened 2017-02-28
ELPP-1459 Fixed bugs in article fragment support.

jamiehollern:
@giorgiosironi @nlisgo This now works having tested locally.

Nb Lax rejects non-HTTPS URLs so this won't work locally without a slight change to the URL path of the images.


giorgiosironi:
Will merge after end2end is green, wait.

PR #102, opened 2017-03-02
ELPP-2341 Added caching to REST endpoints.

jamiehollern:
@nlisgo @giorgiosironi Completed rest endpoint caching work.

### DIFF
--- a/src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
+++ b/src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
@@ -4,13 +4,12 @@ namespace Drupal\jcms_rest\Plugin\rest\resource;
 
 use DateTimeImmutable;
 use DateTimeZone;
-use Drupal\Component\Utility\Random;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\Query\QueryInterface;
 use Drupal\Core\Field\FieldItemListInterface;
 use Drupal\image\Entity\ImageStyle;
-use Drupal\jcms_article\ArticleCrud;
 use Drupal\jcms_rest\Exception\JCMSBadRequestHttpException;
+use Drupal\jcms_rest\PathMediaTypeMapper;
 use Drupal\node\Entity\Node;
 use Drupal\paragraphs\Entity\Paragraph;
 use Drupal\rest\Plugin\ResourceBase;
@@ -389,7 +388,7 @@ abstract class AbstractRestResourceBase extends ResourceBase {
   protected function fieldValueFormatted(FieldItemListInterface $data) {
     $view = $data->view();
     unset($view['#theme']);
-    return render($view);
+    return \Drupal::service('renderer')->renderPlain($view);
   }
 
   /**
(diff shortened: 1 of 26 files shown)


src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
```diff
@@ -540,4 +551,31 @@ public function getEntityQueueItem(EntityInterface $node, FieldItemListInterface
...
 
+  /**
+   * Returns the endpoint from the rest resource "canonical" annotation.
+   *
+   * @return string
+   */
+  function getEndpoint(): string {
+    $r = new \ReflectionClass(static::class);
+    $annotation = $r->getDocComment();
+    preg_match("/\"canonical\" = \"(.+)\"/", $annotation, $endpoint);
```
giorgiosironi: `preg_match()` may not match and return false, I'd check and throw an exception if that's not the case.

jamiehollern: Good shout, thanks.

src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
```diff
@@ -540,4 +539,38 @@ public function getEntityQueueItem(EntityInterface $node, FieldItemListInterface
...
+   * @throws \Exception
+   */
+  public function getContentType(): string {
+    $endpoint = $this->getEndpoint();
+    $mapper = new PathMediaTypeMapper();
+    $content_type = $mapper->getMediaTypeByPath($endpoint);
+    if (!$content_type) {
+      throw new \Exception('Content type not found for specified rest resource.');
+    }
+    return $content_type . ';version=1';
```
nlisgo: We should create a separate ticket to allow better control of the version number to reduce technical debt.

Actually, it would be an improvement to allow the version number to be passed as a parameter with a default value of 1.

`function getContentType($version = 1): string {`

src/modules/jcms_rest/src/Plugin/rest/resource/CommunityListRestResource.php
```diff
@@ -54,7 +55,8 @@ public function get() {
         }
       }
     }
-    $response = new JsonResponse($response_data, Response::HTTP_OK, ['Content-Type' => 'application/vnd.elife.community-list+json;version=1']);
+    $response = new JCMSRestResponse($response_data, Response::HTTP_OK, ['Content-Type' => $this->getContentType()]);
+    $response->addCacheableDependencies($nodes);
```
nlisgo: Where is `nodes[]` populated?

jamiehollern: Line 50.

src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
jamiehollern: I'm happy to do this but when versioning is done properly this will need to change.

src/modules/jcms_rest/src/Plugin/rest/resource/CollectionListRestResource.php
```diff
@@ -55,7 +56,8 @@ public function get() {
         }
       }
     }
-    $response = new JsonResponse($response_data, Response::HTTP_OK, ['Content-Type' => 'application/vnd.elife.collection-list+json;version=1']);
+    $response = new JCMSRestResponse($response_data, Response::HTTP_OK, ['Content-Type' => $this->getContentType()]);
```
giorgiosironi: two lines for each endpoint, making explicit the content type and the caching, good enough for me; I'll leave the rest to @nlisgo

PR #105, opened 2017-03-07
Add reproducibility project cancer biology collection

nlisgo:
Do we need to add in the replication studies also? I have not yet, I will await instructions.


thewilkybarkid:
They should be there too, so the order is 07072, 21253, 06959, 17584 etc.

nlisgo:
@sianroderick who should the curator be for this collection? Each collection requires at least one curator.

sianroderick:
We need all articles present - including the Editorials, Insights and Feature Articles.

I'll check with Andy/Peter to find out who the curator should be.

nlisgo:
@sianroderick @thewilkybarkid all articles (including the Editorials, Insights and Feature Articles) should now be present in the list in the expected order.

We just need to assign a curator. Currently, the curator is not optional so I need to assign one. I suggest we merge this in with an arbitary curator and resolve the issue before going live.

PR #106, opened 2017-03-08
Prepopulate highlight list with random nodes

nlisgo:


src/modules/jcms_migrate/jcms_migrate.module
```diff
@@ -0,0 +1,14 @@
...
+ * JCMS migrate module file.
+ */
+
+use Drupal\Core\Database\Query\AlterableInterface;
+
+/**
+ * Implements hook_query_TAG_alter().
+ */
+function jcms_migrate_query_random_alter(AlterableInterface $query) {
+  $query->orderRandom();
```
jamiehollern: Nice.

PR #107, opened 2017-03-09
ELPP-2362 Updated Lax fragment integration.

jamiehollern:
@nlisgo

### DIFF
--- a/src/modules/jcms_article_fragment/jcms_article_fragment.install
+++ b/src/modules/jcms_article_fragment/jcms_article_fragment.install
@@ -1,0 +1,13 @@
+<?php
+
+/**
+ * @file
+ * Contains jcms_article_fragment install/update functions.
+ */
+
+/**
+ * Implements hook_update_N().
+ */
+function jcms_article_fragment_update_8101() {
+  \Drupal::entityDefinitionUpdateManager()->applyUpdates();
+}
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,4 @@ scripts/legacy_cms_files_alt/
 vendor/
 web/
 build/
-/console/cache
+/console
--- a/src/modules/jcms_article_fragment/jcms_article_fragment.module
+++ b/src/modules/jcms_article_fragment/jcms_article_fragment.module
@@ -6,6 +6,7 @@
  */
 
 use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Form\FormStateInterface;
 
 /**
  * Implements hook_help().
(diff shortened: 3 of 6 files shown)


src/modules/jcms_article_fragment/jcms_article_fragment.install
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * @file
+ * Contains jcms_article_fragment install/update functions.
+ */
+
+/**
+ * Implements hook_update_N().
+ */
+function jcms_article_fragment_update_8001() {
+  \Drupal::entityDefinitionUpdateManager()->applyUpdates();
```
nlisgo: Why is this necessary?

jamiehollern: The article fragment entity schema has changed in this piece of work from the original piece of work. This simply updates it in places where it's already installed.

nlisgo: This should be handled with:

```
drush cim -y
```

which should be run as a deploy step in all environments. While it does no harm to add it, it should not be necessary so please remove.

src/modules/jcms_article_fragment/jcms_article_fragment.install
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * @file
+ * Contains jcms_article_fragment install/update functions.
+ */
+
+/**
+ * Implements hook_update_N().
+ */
+function jcms_article_fragment_update_8001() {
```
nlisgo: Nitpick: should be `8101` I believe.

src/modules/jcms_article_fragment/jcms_article_fragment.install
```diff
@@ -0,0 +1,13 @@
...
+ * @file
+ * Contains jcms_article_fragment install/update functions.
+ */
+
+/**
+ * Implements hook_update_N().
+ */
+function jcms_article_fragment_update_8101() {
+  \Drupal::entityDefinitionUpdateManager()->applyUpdates();
+}
```
jamiehollern: @nlisgo Having double checked this, this hook is necessary. The schema is not stored in configuration but in code so it needs to be updated this way.

src/modules/jcms_article_fragment/jcms_article_fragment.install
jamiehollern: Done.

nlisgo:
I'll merge as soon as travis goes green!

PR #108, opened 2017-03-09
Set community items at random

nlisgo:


nlisgo:
@jamiehollern, there were a couple of minor additional things that I have snuck into this PR. The important commit is the last one.

We will now be able to use `vagrant share` (https://www.vagrantup.com/docs/share/). And one commit gets rid of a bug in the `podcast_episodes.json` fixture.

PR #112, opened 2017-03-13
Add images for subjects and covers

nlisgo:


nlisgo:
After a discussion with the team we are going to remove the cover images from the git repo and put them in a public S3 bucket.

There is concern that if we add them now they will unnecessarily permanently inflate the size of the repo. Since we are going to remove the migration fixtures post launch it makes sense to consider the impact of these files on the git history.

PR #118, opened 2017-03-15
Adding impact statements

sianroderick:


src/modules/jcms_migrate/migration_assets/subjects.json
```diff
@@ -2,105 +2,105 @@
   {
     "id": "plant-biology",
     "name": "Plant Biology",
-    "impact_statement": "Plant Biology impact statement",
+    "impact_statement": "eLife publishes plant biology research including development, epigenetics, genomics, signalling, metabolism and plant–microbe interactions. Decisions are made by <a href="https://beta.elifesciences.org/about/people">journal editors</a> who are active researchers in plant biology.",
```
nlisgo: These links should be relative so we don't need to change the again once we go live:

```
<a href="/about/people">journal editors</a>
```

thewilkybarkid: Hmm, we can't assume that it's Journal reading the API, so they should be absolute, but not at `beta.elifesciences.org` (`https://elifesciences.org` instead). Journal should rewrite these so that they're relative (so following them doesn't take you to a different instance).

PR #119, opened 2017-03-16
Filter events by show parameter

nlisgo:


src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
```diff
@@ -327,12 +329,35 @@ protected function filterPageAndOrder(QueryInterface &$query, $sort_by = NULL) {
...
+      'closed' => 'end-date',
+      'open' => 'start-date',
+    ];
+
+    if (in_array($show_option, array_keys($options))) {
+      self::$requestOptions[$options[$show_option]] = date('Y-m-d');
+      $this->filterDateRange($query, 'field_event_datetime.end_value', FALSE);
+    }
+    elseif ($show_option != 'all') {
+      throw new JCMSBadRequestHttpException(t('Invalid show option'));
```
jamiehollern: This should probably be a `UnexpectedValueException`.

nlisgo: @thewilkybarkid the last time we chatted (3rd feb on slack) about this, I was concerned about consistency. 

It is my understanding that we would return:

"400 Bad Request if any of the passed parameters are not as expected?"

Maybe this is different because it is an enum (`['all', 'open', 'closed']`)?

thewilkybarkid: Yeah, `show=banana` should result in a 400.

PR #125, opened 2017-03-20
Test article bulk import in end2end enviromment

giorgiosironi:


nlisgo:
Have just restarted the build on travis.

Jenkinsfile
```diff
@@ -15,6 +15,7 @@ elifePipeline {
             {
                 builderDeployRevision 'journal-cms--end2end', commit
                 builderSmokeTests 'journal-cms--end2end', '/srv/journal-cms'
+                builderCmd 'journal-cms--end2end', 'cd /srv/journal-cms; drush article-import-all --limit=25'
```
nlisgo: in order for drush to run contextually it should be run from `/srv/journal-cms/web` and we should then reference the `drush` command in the the `vendor/bin` folder.

```
builderCmd 'journal-cms--end2end', 'cd /srv/journal-cms/web; ../vendor/bin/drush article-import-all --limit=25'
```

giorgiosironi: Right :+1:

giorgiosironi:
If we will do this, we don't do it in end2end.

PR #127, opened 2017-03-21
Update community list, events and labs

nlisgo:


nlisgo:
Thanks for the nod @jamiehollern :)

PR #136, opened 2017-03-24
[WIP] Add /ping

giorgiosironi:
Goal: add a /ping API that just returns the text `pong` plus a few "don't cache anything" headers.

### DIFF
--- a/src/modules/jcms_rest/src/Plugin/rest/resource/PingRestResource.php
+++ b/src/modules/jcms_rest/src/Plugin/rest/resource/PingRestResource.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Drupal\jcms_rest\Plugin\rest\resource;
+
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Provides a ping for smoke tests to call.
+ *
+ * @RestResource(
+ *   id = "ping_rest_resource",
+ *   label = @Translation("Ping rest resource"),
+ *   uri_paths = {
+ *     "canonical" = "/ping"
+ *   }
+ * )
+ */
+class PingRestResource extends AbstractRestResourceBase {
+  /**
+   * Responds to GET requests.
+   *
+   * @return Symfony\Component\HttpFoundation\Response
+   */
+  public function get(int $number) {
+    return new Response(200, 'pong'); // TODO: add headers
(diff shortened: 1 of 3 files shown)


src/modules/jcms_rest/src/Plugin/rest/resource/PingRestResource.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @RestResource(
+ *   id = "ping_rest_resource",
+ *   label = @Translation("Ping rest resource"),
+ *   uri_paths = {
+ *     "canonical" = "/ping"
+ *   }
+ * )
+ */
+class PingRestResource extends AbstractRestResourceBase {
```
thewilkybarkid: There must be a simpler way to add a controller than extending whatever `AbstractRestResourceBase` is?

nlisgo: There is a simpler way. But to complicate things slightly I think we should have this in it's own module.

The missing step to get this working is that we need to set the permissions to access the resource. This is typically done in the UI and then we can export the configuration.

@giorgiosironi let's discuss which is the best approach to implement this. Do we want it to always be available or is something we could enable, test, then tear down?

giorgiosironi: Probably, @nlisgo shall I use the @RestResource support or can I go for something simpler like https://www.drupal.org/docs/8/creating-custom-modules/adding-a-basic-controller ? In which module should this be added? Does it need a new one?

nlisgo: I think it needs a new one. Would `jmcs_ping` be an appropriate name?

thewilkybarkid: We'd probably want it on any Drupal, so maybe just `ping`? (Or `elife_ping` due to https://www.drupal.org/project/ping?)

giorgiosironi: always available, we use it in our /status page

giorgiosironi:
Obsolete by #137.

PR #138, opened 2017-03-28
Remove coveralls until we improve coverage

nlisgo:


giorgiosironi:
The smell of green builds in the morning, thanks!

PR #142, opened 2017-04-03
Drush migrate output can be checked by this script

giorgiosironi:


giorgiosironi:
Required by https://github.com/elifesciences/journal-cms-formula/pull/25

PR #160, opened 2017-04-15
Add article metrics for cover list ordering

nlisgo:
See: https://github.com/elifesciences/journal-cms-formula/pull/30
See: https://github.com/elifesciences/builder-private/pull/15

### DIFF
--- a/src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
+++ b/src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
@@ -5,6 +5,7 @@
  * Drush commands for the module.
  */
 
+use Drupal\jcms_article\Entity\ArticleVersions;
 use GuzzleHttp\Exception\RequestException;
 use Drupal\node\Entity\Node;
 use Drupal\taxonomy\Entity\Term;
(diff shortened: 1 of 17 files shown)


giorgiosironi:
I reviewed and merged the changes to the formula and builder-private, they are in now. However, we should first get a green end2end and prod of the project before merging this additional change.

src/modules/jcms_notifications/drush/jcms_notifications.drush.inc
```diff
@@ -98,11 +113,59 @@ function drush_jcms_notifications_article_import_all() {
...
+      '!minutes' => floor($time / 60),
+      '!seconds' => round($time % 60),
+    ]));
+  }
+}
+
+/**
  * Callback function drush_jcms_notifications_article_import().
  */
 function drush_jcms_notifications_article_import($lrp = FALSE) {
```
giorgiosironi: the name of this drush command is outdated, but let's not rename it now as it involves the formula. We'll do that incrementally

nlisgo: Yes, I will do a followup to change it.

PR #161, opened 2017-04-18
Clean database on rollback

giorgiosironi:
The default rollback step will change the version of the code being required and run a new salt highstate. This doesn't play well with the journal-cms database.

Therefore, this custom rollback step cleans the database instead and
reinstalls the older commit, but without existing state getting in the
way. We lose a bit of useful feedback here because end2end is not as
stateful as production, but we gain in stability as code that is broken
does not stay deployed, and other projects may run end2end again
successfully.


giorgiosironi:
Let's see...

PR #170, opened 2017-04-25
Disable Fragment API request if no auth key set

nlisgo:
Allow metrics to be collected for unpublished articles

### DIFF
--- a/src/modules/jcms_article/src/Hooks/NodePresave.php
+++ b/src/modules/jcms_article/src/Hooks/NodePresave.php
@@ -138,6 +138,10 @@ final class NodePresave {
    * @param \Drupal\jcms_article\Entity\ArticleVersions $article
    */
   public function updateFragmentApi(EntityInterface $entity, ArticleVersions $article) {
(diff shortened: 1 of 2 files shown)


src/modules/jcms_article/src/Hooks/NodePresave.php
```diff
@@ -138,6 +138,10 @@ public function setSubjectTerms(EntityInterface $entity, ArticleVersions $articl
    * @param \Drupal\jcms_article\Entity\ArticleVersions $article
    */
   public function updateFragmentApi(EntityInterface $entity, ArticleVersions $article) {
+    if (empty(Settings::get('jcms_article_auth_unpublished', FALSE))) {
+      return;
```
nlisgo: I should only be posting to fragment API if the node was saved in the UI.

nlisgo: This will be address in PR: https://github.com/elifesciences/journal-cms/pull/171

PR #188, opened 2017-05-25
Introduce pagination for highlight lists

nlisgo:
This blocked until api-sdk and journal is updated


nlisgo:
Need to update api-sdk before merging to incorporate relevant changes there to api-sdk and api-client: 

```
composer update elife/api-skd
```

PR #194, opened 2017-05-26
Change labs-experiment to labs-post

nlisgo:
This is just the minimal changes needed. More refactoring will be done post launch


giorgiosironi:
From what I understand this will become green once the api-raml is changed.

nlisgo:
@giorgiosironi that is my understanding. It doesn't surprise me that it failed CI without this update.

giorgiosironi:
Pushed the update to api-raml in the composer.lock, but conflicts with `develop`. Since more than one person is using this branch, I will merge `develop` and update again rather then rebasing, which would be dangerous.

giorgiosironi:
Updated api-raml, but there were two tests failing. The first one was an obvious change, trying to patch the second which seems related to the `type` field coming out of the API.

giorgiosironi:
I am adding commits trying to fix the `type` fields, but they can be dropped or reverted later if they don't help.

giorgiosironi:
Green, but needs the review of @nlisgo

PR #204, opened 2017-06-05
view_unpublish is not necessary

nlisgo:


nlisgo:
I will remove the module from composer in a followup

PR #205, opened 2017-06-05
Monitor exceptions

giorgiosironi:
There are already logging subscribers that append controller exceptions to the Monolog error.json file and to the PHP error log. This adds the exception to New Relic, where it will be counted in the Error Rate and possibly trigger email alerts.


nlisgo:
@giorgiosironi I'm happy with the code just not sure it belongs in the jcms_rest module.

PR #208, opened 2017-06-07
Using HTTPS in smoke tests

giorgiosironi:
Recently (today) the upgrade from HTTP to HTTPS started working, so these URLs when loaded in HTTP will get a 301 redirect to their HTTPS equivalent.

Still, locally or if no ENVIRONMENT_NAME is specified (other VMs), we leave the possibility to use HTTP.

### DIFF
--- a/smoke_tests.sh
+++ b/smoke_tests.sh
@@ -2,7 +2,12 @@
 set -ex
 
 local_hostname=$(hostname)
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Not urgent, will remove the redirect for now.

giorgiosironi:
Obsolete, let's not mess with this.

PR #219, opened 2017-06-22
ELPP-2687 Add label and id fields for image and table in content blocks

nlisgo:


giorgiosironi:
There is a problem with the MySQL version (used in CI, elsewhere we use RDS) being installed, will rebuild it now.

giorgiosironi:
MySQL version is consistently `5.7.17-1ubuntu14.04` everywhere, the problem of the missing socket is not coming from there. Investigating some more.

giorgiosironi:
Not really, new installations have `5.7.18-1ubuntu14.04`, I'll pin that everywhere.

giorgiosironi:
Now green, I'll apply the formula to all environments.

PR #237, opened 2017-07-28
ELPP-3012 Disable internal page cache of journal

nlisgo:


src/modules/jcms_rest/tests/src/Functional/CacheTest.php
```diff
@@ -0,0 +1,106 @@
...
+    $this->assertEquals('MISS', $response->getHeaderLine('X-Drupal-Cache'), 'Page was not cached.');
+    $response = $this->makeGuzzleRequest($endpoint, ['Accept' => 'application/vnd.elife.subject-list+json;version=1']);
+    $this->assertEquals(HttpFoundationResponse::HTTP_OK, $response->getStatusCode());
+    $this->assertEquals('HIT', $response->getHeaderLine('X-Drupal-Cache'), 'Page cached.');
+    $response = $this->makeGuzzleRequest($endpoint, ['Accept' => 'application/vnd.elife.subject-list+json;version=1', 'If-None-Match' => $response->getHeaderLine('Etag')]);
+    $this->assertEquals(HttpFoundationResponse::HTTP_NOT_MODIFIED, $response->getStatusCode());
+    $this->assertEquals('MISS', $response->getHeaderLine('X-Drupal-Cache'), 'Page was not cached.');
+    $response = $this->makeGuzzleRequest($endpoint, ['Accept' => 'application/vnd.elife.subject-list+json;version=1', 'If-None-Match' => $response->getHeaderLine('Etag')]);
+    $this->assertEquals(HttpFoundationResponse::HTTP_NOT_MODIFIED, $response->getStatusCode());
+    $this->assertEquals('HIT', $response->getHeaderLine('X-Drupal-Cache'), 'Page cached.');
```
giorgiosironi: all the tested behavior for the cache seems correct. But wasn't the original specification that backends should not cache responses at all? (because the CDN for the gateway and Journal's internal HTTP cache already do)

PR #244, opened 2017-08-11
ELPP-2588 Add containing parameter to collections

thewilkybarkid:


src/modules/jcms_rest/src/Plugin/rest/resource/CollectionListRestResource.php
```diff
@@ -38,6 +39,26 @@ public function get() {
...
+      foreach ($containing as $item) {
+        preg_match('~^(article|blog-article|interview)/([a-z0-9-]+)$~', $item, $matches);
+
+        if (empty($matches[1]) || empty($matches[2])) {
+          throw new JCMSBadRequestHttpException(t('Invalid containing parameter'));
+        }
+
+        $orCondition = $orCondition->condition($base_query->andConditionGroup()
+          ->condition('field_collection_content.entity.type', str_replace('-', '_', $matches[1]))
+          ->condition('field_collection_content.entity.uuid', $matches[2], 'ENDS_WITH')
```
thewilkybarkid: This looks like how ID matching is done elsewhere, but isn't right: it shouldn't be possible to do partial matching.

thewilkybarkid:
How do I test this?

thewilkybarkid:
Appears we have no way of testing this. /cc @giorgiosironi

src/modules/jcms_rest/src/Plugin/rest/resource/CollectionListRestResource.php
giorgiosironi: is it because the `uuid` is the full one in the database (e.g. `de43e8c6-8196-11e7-bb31-be2e44b06b34`) and we are only using its end (e.g. `44b06b34`)? I wonder how duplicates protection works.

giorgiosironi:
If we add enough of a Shared Fixture, populating the database, we can extend the API tests to make a fixed request and expect a fixed response?

thewilkybarkid:
Just spotted this won't actually work with articles, since they don't (mis)use the UUID.

PR #245, opened 2017-08-11
Upgrade drupal-vm

nlisgo:


giorgiosironi:
Is it related to the case insensitive filesystem issue?

nlisgo:
Yes. I thought before I raised an issue on the `drupal-vm` project I should update to the latest version and confirm it is a problem.

PR #246, opened 2017-08-14
Fix inline image blocks

thewilkybarkid:
There's currently a problem where the `field_block_image_inline` field doesn't exists. This clearly is a bigger issue, but we can unbreak API responses by checking for it.


thewilkybarkid:
No tests around this. No idea where to start with that.

giorgiosironi:
https://github.com/elifesciences/journal-cms/blob/develop/src/modules/jcms_rest/tests/src/Functional/RecursiveEndpointValidatorTest.php should in principle hit this and it doesn't?

thewilkybarkid:
Looks to me like {{RecursiveEndpointValidatorTest}} is hitting the locally installed site rather than fixtures?

thewilkybarkid:
Disabling these changes on a copy of the production DB shows the test failures. Adding the changes shows the tests passing.

Clearly this way of testing isn't right.

thewilkybarkid:
Merging, as it seems to work.

PR #253, opened 2017-09-01
ELPP-3129 Improve logging on fetch of article versions and metrics

nlisgo:


src/modules/jcms_article/src/FetchArticleMetrics.php
```diff
@@ -53,11 +54,16 @@ public function setEndpoint($endpoint) {
...
    * @return \Drupal\jcms_article\Entity\ArticleMetrics
+   * @throws \InvalidArgumentException
    * @throws \TypeError
    */
   public function getArticleMetrics(string $id): ArticleMetrics {
     $response = $this->requestArticleMetrics($id);
-    $json = json_decode((string) $response->getBody(), TRUE);
+    $json = (string) $response->getBody() ?: '{}';
+    $json = json_decode($json, TRUE);
+    if (json_last_error() !== JSON_ERROR_NONE) {
```
thewilkybarkid: Use Guzzle's `json_decode`, it throws an exception for you.

src/modules/jcms_article/src/FetchArticleVersions.php
```diff
@@ -68,24 +70,36 @@ public function getArticleVersions(string $id): ArticleVersions {
...
+      $response = $this->client->get($url, $options);
+      if ($response instanceof ResponseInterface) {
+        \Drupal::logger('jcms_article')
+          ->notice(
+            'Article versions have been requested @url with the response: @response',
+            ['@url' => $url, '@response' => \GuzzleHttp\Psr7\str($response)]
+          );
+        return $response;
+      }
+      throw new \TypeError('Network connection interrupted on request.');
```
thewilkybarkid: It's not a `TypeError`.

nlisgo: I'm thinking we can just remove this. If `ResponsInterface` is not returned then an exception will be thrown by guzzle and it can certainly handle lose of connection.

PR #254, opened 2017-09-18
Revamping API tests on canned data

giorgiosironi:
- Tests for 404 on item are working
- Checking the data generation script is successful
- Added all content types that already pass tests. The randomized data is validating with the current schema for these.

### DIFF
--- a/src/modules/jcms_rest/tests/src/Functional/RamlSchemaValidationTest.php
+++ b/src/modules/jcms_rest/tests/src/Functional/RamlSchemaValidationTest.php
@@ -107,6 +107,65 @@ class RamlSchemaValidationTest extends UnitTestCase {
         'application/vnd.elife.event-list+json;version=1',
         'application/vnd.elife.event+json;version=1',
       ],
+      /*
+       * fails because there is no `chapters` property
+      [
+        'GET',
+        '/podcast-episodes',
+        'number',
+        'application/vnd.elife.podcast-episode-list+json;version=1',
+        'application/vnd.elife.podcast-episode+json;version=1',
+      ],
+       */
+
+      [
+        'GET',
+        '/interviews',
+        'id',
+        'application/vnd.elife.interview-list+json;version=1',
+        'application/vnd.elife.interview+json;version=1',
+      ],
+      /*
+       * fails because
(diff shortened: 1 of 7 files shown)


giorgiosironi:
Locally the uncommented tests passed, but on ci they don't; probably a difference in content, resetting it locally.

giorgiosironi:
Things like `items[0].selectedCurator.orcid` need to be customized to generate a correct ORCID.

src/modules/jcms_rest/tests/src/Functional/RamlSchemaValidationTest.php
```diff
@@ -12,30 +12,47 @@
...
+      self::$contentGenerated = TRUE;
+      $projectRoot = realpath(__DIR__ . '/../../../../../..');
+      $script = $projectRoot . '/scripts/generate_content.sh';
+      if (!file_exists($script)) {
+        throw new RuntimeException("File $script does not exist");
+      }
+      $logFile = '/tmp/generate_content.log';
+      exec("$script >$logFile 2>&1", $output, $exitCode);
+      if ($exitCode != 0) {
+        throw new RuntimeException("$script failed. Check log file $logFile");
```
thewilkybarkid: `symfony/process` is available, can simplify this.

sync/field.field.node.annual_report.field_annual_report_year.yml
```diff
@@ -17,7 +17,9 @@ default_value: {  }
 default_value_callback: ''
 settings:
   min: 2012
-  max: null
+  # default maximum value is 999 in:
+  # ./web/core/lib/Drupal/Core/Field/Plugin/Field/FieldType/IntegerItem.php
+  max: 2099
```
nlisgo: Is this prophetic?

src/modules/jcms_rest/tests/src/Functional/RamlSchemaValidationTest.php
giorgiosironi: It is not in the explicit requirements in composer.json, so I would avoid introducing additional dependencies in place of 3-4 lines of standard library calls

sync/field.field.node.annual_report.field_annual_report_year.yml
giorgiosironi: Space: 2099

PR #256, opened 2017-09-18
ELPP 3160 implement jobs advert api

davidcmoulton:
In Drupal terms, the job advert closing date is currently a `Timestamp`. But could be made a `Date` with minimal rework if that's a better choice.


src/modules/jcms_rest/src/Plugin/rest/resource/JobAdvertListRestResource.php
```diff
@@ -0,0 +1,85 @@
...
+    $response_data = [
+      'total' => 0,
+      'items' => [],
+    ];
+    $nodes = [];
+    if ($total = $count_query->count()->execute()) {
+      $response_data['total'] = (int) $total;
+      $this->filterPageAndOrder($items_query);
+      $nids = $items_query->execute();
+      $nodes = Node::loadMultiple($nids);
```
giorgiosironi: I assume this results in 3 queries:
- count
- list ids
- load all these `$nids`

giorgiosironi: Is there a way to log that? We can do it locally with MySQL's log

src/modules/jcms_rest/src/Plugin/rest/resource/JobAdvertListRestResource.php
```diff
@@ -0,0 +1,85 @@
...
+      if (!empty($nodes)) {
+        foreach ($nodes as $node) {
+          $response_data['items'][] = $this->getItem($node);
+        }
+      }
+    }
+    $response = new JCMSRestResponse($response_data, Response::HTTP_OK, ['Content-Type' => $this->getContentType()]);
+    $response->addCacheableDependencies($nodes);
```
giorgiosironi: does this mean `invalidate cache if one of these nodes is updated`?

src/modules/jcms_rest/src/Plugin/rest/resource/JobAdvertListRestResource.php
```diff
@@ -0,0 +1,85 @@
...
+   *
+   * @return array
+   */
+  public function getItem(EntityInterface $node) {
+    /* @var Node $node */
+    $item = $this->processDefault($node);
+
+    // Impact statement is optional.
+    if ($node->get('field_impact_statement')->count()) {
+      $item['impactStatement'] = $this->fieldValueFormatted($node->get('field_impact_statement'));
```
giorgiosironi: looks duplicated with the item resource. We could share the code snippet with a small `JobAdvertPresenter` (Java-like name). Perhaps it is even more general and usable from multiple content types

sync/field.storage.node.field_job_advert_respons.yml
```diff
@@ -0,0 +1,23 @@
...
+status: true
+dependencies:
+  module:
+    - field_permissions
+    - node
+    - text
+third_party_settings:
+  field_permissions:
+    permission_type: public
+id: node.field_job_advert_respons
```
giorgiosironi: is `respons` as opposed to `response` intentional?

davidcmoulton: It's due to the character limit constraint on the name of the field.

src/modules/jcms_rest/src/Plugin/rest/resource/JobAdvertListRestResource.php
nlisgo: That is what is supposed to happen. But we seem to be having issues with that in production. Disabling cache entirely is the priority so I don't want to spend any more time working why the cache isn't behaving as expected.

giorgiosironi: No problem, just looking to understand more

sync/field.storage.node.field_job_advert_respons.yml
nlisgo: The longer name we were going for was `field_job_advert_responsibilities` which was just 1 character too long

src/modules/jcms_rest/src/Plugin/rest/resource/JobAdvertListRestResource.php
nlisgo: @giorgiosironi there are some redundancies here but when I get around to addressing the slow queries I can do that clean up then. The implementation here is consistent with that of other rest resources.

src/modules/jcms_rest/src/Plugin/rest/resource/JobAdvertListRestResource.php
```diff
@@ -0,0 +1,85 @@
...
+   *
+   * @todo - elife - nlisgo - Handle version specific requests
+   */
+  public function get() {
+    $base_query = \Drupal::entityQuery('node')
+      ->condition('status', NodeInterface::PUBLISHED)
+      ->condition('changed', \Drupal::time()->getRequestTime(), '<')
+      ->condition('type', 'job_advert');
+
+    $this->filterShow($base_query);
```
nlisgo: The `filterShow` method will need some work so it can work for `events` and `job_adverts`. Currently, it is only working for `events`.

giorgiosironi:
Note to self: the generation process is not adding a value for `closingDate`, which is what is probably causing the test failure.

src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
```diff
@@ -359,9 +360,14 @@ protected function filterPageAndOrder(QueryInterface &$query, $sort_by = NULL) {
    * Apply filter by show parameter: all, open or closed.
    *
    * @param \Drupal\Core\Entity\Query\QueryInterface $query
-   * @param string
+   * @param string $filterFieldName The name of the field to filter on
+   * @param bool $isTimeStamp whether $field is a Timestamp field
+   * @throws InvalidArgumentException if the filter field name argument is not supplied
    */
-  protected function filterShow(QueryInterface &$query) {
+  protected function filterShow(QueryInterface &$query, string $filterFieldName = '', bool $isTimeStamp = FALSE) {
```
giorgiosironi: not a big problem for now but I'll keep in mind http://www.informit.com/articles/article.aspx?p=1392524 as this class expands

src/modules/jcms_rest/src/Plugin/rest/resource/JobAdvertItemRestResource.php
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * @param \Drupal\core\Field\FieldItemListInterface $field
+   * @param string $fieldLabel
+   * @param boolean $isSection
+   * @return array
+   */
+  public function getFieldJson($field, $fieldLabel = '', $isSection = FALSE) {
+    $texts = $this->splitParagraphs($this->fieldValueFormatted($field, FALSE));
+    if ($isSection) {
+      return JobAdvertItemRestResource::getFieldJsonAsSection($fieldLabel, $texts);
```
giorgiosironi: you can use `self` in place of the class name

src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
nlisgo: If a value is expected for `$filterFieldName` then we should not set a default value of `''`.

Suggested change:
```
protected function filterShow(QueryInterface &$query, string $filterFieldName, bool $isTimeStamp = FALSE) {
```

PR #257, opened 2017-09-20
Generate random orcid id’s for dummy content

nlisgo:


src/modules/jcms_rest/jcms_rest.services.yml
```diff
@@ -20,3 +20,6 @@ services:
     class: Drupal\jcms_rest\PathMediaTypeMapper
   jcms_rest.monitoring:
     class: eLife\Logging\Monitoring
+  plugin.manager.field.field_type:
+    class: Drupal\jcms_rest\FieldTypePluginManager
+    arguments: ['@container.namespaces', '@cache.discovery', '@module_handler', '@typed_data_manager']
```
giorgiosironi: no idea what these arguments are, but I guess is the constructor that we are extending. Would be nice to leave here a link to the place where this list comes from

src/modules/jcms_rest/src/FieldTypePluginManager.php
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * {@inheritdoc}
+ */
+class FieldTypePluginManager extends FieldTypePluginManagerExtend {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getPluginClass($type) {
```
giorgiosironi: ok, this is the factory essentially

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/StringItem.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+class StringItem extends StringItemExtend {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function generateSampleValue(FieldDefinitionInterface $field_definition) {
+    switch ($field_definition->getName()) {
+      // Random values for orcid need to match the regex ^[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{3}[0-9X]$.
+      case 'field_person_orcid':
```
giorgiosironi: where can a list of these fully qualified field names be found? Is there a table or an admin page that would list them or purely by convention?

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/StringItem.php
```diff
@@ -0,0 +1,33 @@
...
+        $orcid = [4, 4, 4, 3];
+        array_walk($orcid, function (&$value) {
```
giorgiosironi: array_map is usually easier to use as it doesn't modify the original array, but it's functionally the same here

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/StringItem.php
```diff
@@ -0,0 +1,33 @@
...
+          $value = str_pad(rand(0, pow(10, $value-1)-1), $value, '0', STR_PAD_LEFT);
```
giorgiosironi: if the intention is `rand(0, 9999)` shouldn't this be `rand(0, pow(10, $value)-1)`?

nlisgo: Haha, that would explain why all my orcid's look like this:

```
0263-0960-0436-0243
0709-0135-0796-0678
0122-0915-0789-0534
0605-0983-0878-0911
0315-0248-0532-0196
0827-0698-0620-057X
0406-0695-0047-0184
0185-0333-0212-016X
0904-0019-0611-0614
0616-0510-0763-0644
0708-0073-0650-0430
0956-0685-0025-0054
0102-0335-0069-0452
0760-0024-0937-0380
0172-0457-0541-0165
0076-0079-0816-0469
0210-0268-0934-0498
0857-0649-0846-0815
0045-0960-0608-0754
0544-0656-0833-0435
0991-0551-0559-0097
0586-0965-0462-0969
0316-0073-0195-0987
0176-0541-0829-0913
0413-0737-0389-0379
```

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/StringItem.php
nlisgo: The fields each have their own database tables but you can find them in the admin UI of `journal-cms`. Either as a field against the content type (e.g. `/admin/structure/types/manage/annual_report/fields`) or as a field against a paragraph type (e.g. `/admin/structure/paragraphs_type/affiliation/fields`)

src/modules/jcms_rest/jcms_rest.services.yml
nlisgo: These are the arguments required for the `FieldTypePluginManager`. This is just a way for me to override the Drupal core `core.services.yml` entry and intercept the `getPluginClass` method.

See: https://github.com/drupal/drupal/blob/8.3.7/core/core.services.yml#L603-L605

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/StringItem.php
nlisgo: Will take a look. Would you prefer to use a library that generated a value based upon the regex?

giorgiosironi: No need, 5 lines of code are very understandable

src/modules/jcms_rest/src/FieldTypePluginManager.php
nlisgo: Yes.

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/StringItem.php
```diff
@@ -0,0 +1,35 @@
...
+  public static function generateSampleValue(FieldDefinitionInterface $field_definition) {
+    switch ($field_definition->getName()) {
+      // Random values for orcid need to match the regex ^[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{3}[0-9X]$.
+      case 'field_person_orcid':
+        // Generate random string 0000-0000-0000-0000.
+        $orcid = implode('-', array_map(function ($value) {
+          return str_pad(rand(0, pow(10, $value)-1), $value, '0', STR_PAD_LEFT);
+        }, array_fill(0, 4, 4)));
+        // Some orcid's may have a last character of X.
+        $orcid = (rand(0, 10) === 0) ? substr($orcid, 0, -1) . 'X' : $orcid;
```
giorgiosironi: the last symbol would be a checksum apparently, but we don't validate that

nlisgo: So, you think I should just leave the last character as an integer?

thewilkybarkid: We don't check it anywhere, so probably safe to just generate it.

(Though it's easy to do properly and calculate - https://support.orcid.org/knowledgebase/articles/116780-structure-of-the-orcid-identifier. If we do that, we should also generate in the range 0000-0001-5000-0007 to 0000-0003-5000-0001.)

PR #261, opened 2017-11-07
Bump versions of some of the api’s

nlisgo:


scripts/generate_content.sh
```diff
@@ -5,12 +5,12 @@ cd ./web
...
 
 echo "Creating content type subjects"
-../vendor/bin/drush --nocolor generate-terms subjects 5 --kill
+../vendor/bin/drush --nocolor generate-terms subjects 20 --kill
 
 # labs_experiment is actually published as /labs-posts
 for type in blog_article labs_experiment person event podcast_episode interview collection cover press_package annual_report job_advert; do
     echo "Creating content type $type"
-    ../vendor/bin/drush --nocolor generate-content 5 --types=$type
+    ../vendor/bin/drush --nocolor generate-content 5 --types=$type --kill
```
giorgiosironi: this deletes the previously existing content, ok

src/modules/jcms_rest/src/FieldTypePluginManager.php
```diff
@@ -24,6 +35,10 @@ public function getPluginClass($type) {
       // Override class for TimestampItem.
       $plugin_class = TimestampItem::class;
     }
+    if ($type == 'telephone') {
+      // Override class for TelephoneItem.
+      $plugin_class = TelephoneItem::class;
+    }
```
giorgiosironi: this configuration seems consistent with the existing cases, ok

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/EntityReferenceItem.php
```diff
@@ -0,0 +1,35 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public static function generateSampleValue(FieldDefinitionInterface $field_definition) {
+    $values = parent::generateSampleValue($field_definition);
+    switch ($field_definition->getName()) {
+      // Random values must be unique.
+      case 'field_subjects':
+        static $unique = [];
+        // 25 subject terms are generated and the maximum number associated with content is 3.
```
giorgiosironi: not sure what the comment says, the behavior of the class seems to be that we get to up to 10 items in `$unique`, then we only keep the last 2 and go on. The return value seems to be only 1 target.

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/IntegerItem.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+class IntegerItem extends IntegerItemExtend {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function generateSampleValue(FieldDefinitionInterface $field_definition) {
+    $values = parent::generateSampleValue($field_definition);
+    switch ($field_definition->getName()) {
+      // Random values must be unique.
```
giorgiosironi: due to some `unique` database constraint?

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/IntegerItem.php
```diff
@@ -0,0 +1,31 @@
...
+      case 'field_annual_report_year':
+        static $unique = [];
+        while (in_array($values['value'], $unique)) {
+          $values = parent::generateSampleValue($field_definition);
```
giorgiosironi: pay attention to the `minmum` constraint https://github.com/elifesciences/api-raml/blob/67820ba2054346ab47f2a492b8e9f9755aad8c48/src/model/annual-report.v1.yaml#L8

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/TelephoneItem.php
```diff
@@ -0,0 +1,26 @@
...
+class TelephoneItem extends TelephoneItemExtend {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function generateSampleValue(FieldDefinitionInterface $field_definition) {
+    $values = parent::generateSampleValue($field_definition);
+    switch ($field_definition->getName()) {
+      case 'field_block_phone_number':
+          $values['value'] = '+'.$values['value'];
```
giorgiosironi: seems to have a minimum length, if the api-raml reference is correct https://github.com/elifesciences/api-raml/blob/fdc4dc6d2a52db93fb332d412569f932385ba6ea/src/misc/phone-number.v1.yaml#L10

scripts/generate_content.sh
nlisgo: Yes, I thought that would be cleaner?

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/TelephoneItem.php
nlisgo: The value returned by the parent will be a rand number between:
100000000 and 999999999

All this override is doing is prefixing a `+` to that value.

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/IntegerItem.php
nlisgo: The parent method will consider the minimum value of `2012` and also the maximum which is set in the configuration.

See: https://github.com/elifesciences/journal-cms/blob/cc3d7eba87d2d8ea81f012c59f5371d318d4158b/sync/field.field.node.annual_report.field_annual_report_year.yml#L19-L22

src/modules/jcms_rest/src/Plugin/Field/FieldTypeExtend/IntegerItem.php
nlisgo: Correct, the year is the identifier for the `annual-report`. I'm not sure if the value needs to be unique to pass the validation. But, it should be unique.

PR #262, opened 2017-11-22
Add processed field to improve API performance

Dippers:
A new field called field_processed_json has been added to six content types that is used to store static content for each node. This field is populated by a preprocessor using hook_entity_presave. Once populated the processed output is available for rendering the API response making the API response build time much faster.

The affected content types are blog_article, event, interview, labs_experiment and press_package where the paragraph field field_content is processed and stored in field_processed_json. The other content type affected is person where the paragraph fields field_person_profile, field_research_details and field_person_affiliation are processed and stored in field_processed_json.

An update hook is provided to resave all affected nodes that causes the preprocessor to run and populate the new field.

### DIFF
--- a/src/modules/jcms_rest/jcms_rest.module
+++ b/src/modules/jcms_rest/jcms_rest.module
@@ -1,0 +1,30 @@
+<?php
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Implements hook_entity_presave().
+ * 
+ * Preprocess output fields and save in field_processed_json on
+ * node to improve API performance.
+ * 
+ * @param EntityInterface $entity
+ */
+function jcms_rest_entity_presave(EntityInterface $entity) {
+  if ($entity->getEntityType()->id() == 'node'){
+    if ($entity->hasField('field_content') && $entity->hasField('field_processed_json')) {
+      $pluginManager = \Drupal::service('plugin.manager.rest');
+      $processed = $pluginManager->createInstance('blog_article_item_rest_resource', [])->processFieldContent($entity->get('field_content'));
+      $entity->field_processed_json = json_encode($processed);
+    }
+    else {
+      $bundle = $entity->getType();
+      if ($bundle == 'person' && $entity->hasField('field_processed_json')) {
+        $pluginManager = \Drupal::service('plugin.manager.rest');
+        $processed = $pluginManager->createInstance('person_item_rest_resource', [])->getItem($entity);
+        $entity->field_processed_json = json_encode($processed);
+      }
+    }
+  }
+}
--- a/src/modules/jcms_rest/src/Plugin/rest/resource/BlogArticleItemRestResource.php
+++ b/src/modules/jcms_rest/src/Plugin/rest/resource/BlogArticleItemRestResource.php
@@ -64,8 +64,12 @@ class BlogArticleItemRestResource extends AbstractRestResourceBase {
         }
       }
 
-      if ($content = $this->processFieldContent($node->get('field_content'))) {
-        $response['content'] = $content;
+      if ($node->hasField('field_processed_json')) {
+        $processed = $node->get('field_processed_json')->getValue();
+        $response['content'] = json_decode($processed[0]['value']);
+      }
+      else {
+        throw new \Exception("Processed json field not found on entity");
       }
 
       // Subjects is optional.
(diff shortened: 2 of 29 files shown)


src/modules/jcms_rest/src/Plugin/rest/resource/BlogArticleItemRestResource.php
```diff
@@ -64,8 +64,12 @@ public function get($id) {
         }
       }
 
-      if ($content = $this->processFieldContent($node->get('field_content'))) {
-        $response['content'] = $content;
+      if ($node->hasField('field_processed_json')) {
```
giorgiosironi: I like the idea. I am not aware of anything particularly dynamic inside the JSON, so the inputs would change only when editing the item from the interface.

src/modules/jcms_rest/src/Plugin/rest/resource/EventItemRestResource.php
```diff
@@ -67,8 +67,14 @@ public function get($id) {
         $response['uri'] = $node->get('field_event_uri')->first()->getValue()['uri'];
       }
       // Content is optional, only display if there is no Event URI.
-      elseif ($content = $this->processFieldContent($node->get('field_content'))) {
-        $response['content'] = $content;
+      else{
```
giorgiosironi: this logic seems duplicated in all the resources, can it be extracted into a base class or a trait or similar?

src/modules/jcms_rest/jcms_rest.module
```diff
@@ -0,0 +1,30 @@
...
+ * Preprocess output fields and save in field_processed_json on
+ * node to improve API performance.
+ * 
+ * @param EntityInterface $entity
+ */
+function jcms_rest_entity_presave(EntityInterface $entity) {
+  if ($entity->getEntityType()->id() == 'node'){
+    if ($entity->hasField('field_content') && $entity->hasField('field_processed_json')) {
+      $pluginManager = \Drupal::service('plugin.manager.rest');
+      $processed = $pluginManager->createInstance('blog_article_item_rest_resource', [])->processFieldContent($entity->get('field_content'));
```
giorgiosironi: this line seems hardcoded to `blog article`. Is it intentional?

giorgiosironi: if so, the `if()` condition can be made more clear?

sync/field.field.node.event.field_processed_json.yml
```diff
@@ -0,0 +1,19 @@
...
+entity_type: node
+bundle: event
+label: 'Processed (JSON)'
+description: ''
+required: false
+translatable: true
+default_value: {  }
+default_value_callback: ''
+settings: {  }
+field_type: string_long
```
giorgiosironi: I'm not sure how does this get mapped at the MySQL level, but should it be a JSON column (we are on 5.7+)?

src/modules/jcms_rest/jcms_rest.module
Dippers: The problem here is accessing a class method from a hook. To do this a class needs to be instantiated that uses AbstractRestResourceBase as its base class but it does not matter which one so blog_article_item_rest_resource was chosen. An alternative would be to make processFieldContent() a static method but this would have wider implications.

sync/field.field.node.event.field_processed_json.yml
Dippers: This is a Drupal field so we can only use the available field types. If we were using a custom schema we might be able to use JSON column type.

src/modules/jcms_rest/src/Plugin/rest/resource/EventItemRestResource.php
Dippers: We could make it a trait but it is only 6 lines.

giorgiosironi: Multiplied by the times it is repeated, it's 36 lines :)
Most importantly, should be easy to reuse in new kinds of entities and packaging it up in a trait makes that easier

src/modules/jcms_rest/jcms_rest.module
giorgiosironi: Seems to use references to `$this` so I agree on the wider surgery required. Extract a `any_abstract_rest_resource()` function that instantiates this, with this comment as a docblock to communicate this?

nlisgo:
Closing in favour of: https://github.com/elifesciences/journal-cms/pull/262

PR #263, opened 2017-12-06
Add processed fields to improve API performance

nlisgo:
This draws on the approach in: https://github.com/elifesciences/journal-cms/pull/262

There are 4 fields for which we want to store the processed values to improve the read speeds when responding to a pull request.

1. `field_content` - this is a complex field on multiple content types (`blog-articles`, `events`, `interviews`, `labs-posts` and `press-packages`).
1. `field_person_affiliation` - this is stored against the `people` content type.
1. `field_person_profile` - this is stored against the `people` content type.
1. `field_research_details` - this is stored against the `people` content type.

This pull request will create the new fields and then populate a processed value for all of the affected content.

Each time we edit content we will re-evaluate the processed field values.


giorgiosironi:
I can't tell from the configuration, but does this branch turn off the response cache? To do correct performance testing we need to do that, in order to get the real response times we will eventually have in production.

sync/core.extension.yml
```diff
@@ -18,7 +18,6 @@ module:
   datetime_range: 0
   devel_generate: 0
   diff: 0
-  dynamic_page_cache: 0
   editor: 0
```
nlisgo: This module is responsible for cache.

sync/core.extension.yml
```diff
@@ -50,7 +49,6 @@ module:
   monolog: 0
   node: 0
   options: 0
-  page_cache: 0
```
nlisgo: This module is also responsible for cache. Removing this configuration will uninstall the modules from the production instance.

nlisgo:
@giorgiosironi I've commented on the lines in the pull request that cause caching to be disabled.

PR #264, opened 2017-12-11
Lock titles for content that may have annotations

nlisgo:
If the slug of the following content changes then annotations that have been made on the content may become disassociated with that content:

* blog-article
* interview
* labs-post
* press-package

The annotations service should be extended to reassign annotations to a new target if the url changes. Until then we are locking the slug field (title and preferred name for the above content) so only administrators can edit it. 

If an editor is wishing to edit it then they most contact an administrator. The administrator will check to see if the content has any annotations assigned to it, if not then we can change the slug field otherwise the annotations should be altered via the Hypothesis API.

### DIFF
--- a/src/modules/jcms_admin/jcms_admin.module
+++ b/src/modules/jcms_admin/jcms_admin.module
@@ -8,6 +8,7 @@ use Drupal\Core\Database\Query\Condition;
 use Drupal\Core\Database\Query\SelectInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Field\FieldFilteredMarkup;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Site\Settings;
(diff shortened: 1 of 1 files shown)


giorgiosironi:
It's a sensible precaution to take.

We should consider in the future to provide a slug value as part of the API response, making it a first class citizen. Here I think we are only targeting the source field that clients like `journal` use to generate the slug.

Don't know the cost of it, but it could be helpful if people often rewrite their titles: the value could be locked only after publishing rather than immediately (if that's the current intention).

nlisgo:
@giorgiosironi only the edit form is affected. An editor will still be able to set a `title` or `preferred name` for new content.

thewilkybarkid:
Hang on, maybe we can use `shortlink` - https://github.com/hypothesis/client/blob/c8737ff229360b6909687dae74deb9bf19dfb430/src/annotator/plugin/document.coffee#L130

giorgiosironi:
The change seems easily reversible, and it's more conservative in not allowing a dangerous behavior. I'd say go ahead for now.

PR #265, opened 2017-12-15
ELPP-3341 Tolerate multiple X-Consumer-Groups

nlisgo:


src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
```diff
@@ -701,8 +702,8 @@ public function viewUnpublished() {
 
     if (is_null($view_unpublished)) {
       $request = \Drupal::request();
-      $consumer = $request->headers->get('X-Consumer-Groups', 'user');
-      $view_unpublished = ($consumer == 'admin');
+      $consumer = normalize_header($request->headers->get('X-Consumer-Groups', 'user'));
```
giorgiosironi: as the name says, this is not a consumer, but its `groups`

src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
```diff
@@ -701,8 +702,8 @@ public function viewUnpublished() {
...
+      $view_unpublished = (in_array('admin', $consumer));
```
giorgiosironi: outer `(...)` should be optional

src/modules/jcms_rest/src/Response/JCMSRestResponse.php
```diff
@@ -37,9 +38,10 @@ public function addCacheableDependencies(array $dependencies) {
    */
   public function addDefaultCacheableDependencies() {
     $request = \Drupal::request();
-    $consumer = $request->headers->get('X-Consumer-Groups', 'user');
+    $consumer = normalize_header($request->headers->get('X-Consumer-Groups', 'user'));
+    $view_unpublished = (in_array('admin', $consumer));
     $this->setVary('Accept');
-    $max_age = ($consumer == 'admin') ? 0 : Settings::get('jcms_rest_cache_max_age', Cache::PERMANENT);
+    $max_age = ($view_unpublished) ? 0 : Settings::get('jcms_rest_cache_max_age', Cache::PERMANENT);
```
giorgiosironi: also this `(...)` around `$view_unpublished` should be unnecessary

giorgiosironi:
How this was done in Lax: https://github.com/elifesciences/lax/blob/fa2b4a55a7475dafec83b17d88e48bb9b5e3fe87/src/core/middleware.py#L33

PR #268, opened 2018-02-12
Update dependencies

nlisgo:


nlisgo:
Addressing issues with requirement for PHP 7.1 with some of the dependencies.

composer.json
```diff
@@ -19,20 +19,19 @@
...
+    "drupal/console": "^1.0.2",
     "drupal/admin_toolbar": "~1.0",
     "drupal/config_devel": "~1.0-rc1",
     "drupal/config_installer": "~1",
     "drupal/entity_reference_revisions": "1.x-dev",
     "drupal/field_collection": "~1.0-alpha1",
     "drupal/inline_entity_form": "~1.0-beta1",
     "drupal/jsonb": "~1.0-beta1",
-    "drupal/paragraphs": "1.x-dev",
-    "drupal-composer/drupal-project": "8.x-dev",
```
nlisgo: This project is not really setup to be a dependency and is more of a skeleton from drupal projects. I have moved the small amount of code that we were drawing on into `scripts/composer/ScriptHandler.php`.

src/modules/jcms_admin/jcms_admin.info.yml
```diff
@@ -3,6 +3,5 @@ type: module
 description: Admin improvements.
 core: 8.x
 dependencies:
-  - content_moderation
```
nlisgo: Removing this dependency allows me to uninstall the content_moderation module in an update hook. The module is quickly reinstalled when the `../vendor/bin/drush config-import --yes` command is run. As long as we deploy this when the covers preview queue is in the same state as the covers production queue then this should have not noticeable impact.

PR #271, opened 2018-02-14
Paragraph revisions

Dippers:
This PR creates a new drush command that deletes paragraph revisions that are no longer needed. Early versions of the paragraph module created a new paragraph revision every time a node_save was executed. This has lead to more than 1.1 million paragraph revisions being held in the journal-cms database, mainly in the field_article_json field on the article content type.

This drush command identifies all paragraphs revisions associated with a particular field that are not the default or current revision. It then deletes all or some of these revisions as requested.

The available commands are:
drush paragraphs-revisions-delete field_article_json - delete all revisions associated with field_article_json
drush paragraphs-revisions-delete  - delete all revisions on all fields
drush paragraphs-revisions-delete field_article_json --limit=100 - delete up to 100 revisions of field_article_json
drush paragraphs-revisions-delete field_article_json --limit=-1 - just report number of revisions of field_article_json

Deletions are done in blocks of 100 with a progress report at the end of each block.


giorgiosironi:
> Early versions of the paragraph module created a new paragraph revision every time a node_save was executed. 

Does this mean we already have a version in production that does not store too many revisions? I'll cross-check that the database space is not shrinking anymore.

giorgiosironi:
Not sure of that yet:

Dippers:
Yes, the fix was made in the paragraphs module to stop excessive revisions in May 2016 although I'm not sure when you actually deployed it. This clean-up should free up around 6G of database space.

Dippers:
The table optimization will be re-factored so it happens at the end of every call.

nlisgo:
I've merged changes from the upstream and broken this into 2 commands. One for the purge of revisions and the other to optimise the tables.

All of this code will be removed after we perform the purge and optimisation on `prod--journal-cms`.

nlisgo:
I will run this over the weekend on my machine and then recommend that we run it on an instance of `journal-cms` on aws early next week.

giorgiosironi:
Ok, we can do the usual pattern of creating an instance as similar as possible to `continuumtest` and restore the database.

nlisgo:
The output from a local run of `drush pr-purge --feedback=500`:

```
Number of revisions to be deleted is 1680444.
Number of revisions deleted is 500.
Number of revisions deleted is 1000.
Number of revisions deleted is 1500.
Number of revisions deleted is 2000.
Number of revisions deleted is 2500.
...
Number of revisions deleted is 1678500.
Number of revisions deleted is 1679000.
Number of revisions deleted is 1679500.
Number of revisions deleted is 1680000.
Total number of revisions deleted is 1680444.
Success! All revisions for all fields have been purged.

real	244m37.753s
user	46m23.748s
sys	15m10.392s
```

nlisgo:
The output from a subsequent run of `drush pr-optimise`:

```
Optimise paragraph revision tables.
All paragraph revision tables have been optimised.

real	0m5.748s
user	0m0.260s
sys	0m0.164s
```

nlisgo:
After purging all of the redundant paragraph revisions and optimising the paragraph revision tables the uncompressed database dump was 83.2mb, that is down from 8.42gb.

nlisgo:
I've reviewed the affected tables locally and I'm happy to proceeded with a test on a `continuumtestesque` environment. It will feel so good to deploy this.

giorgiosironi:
I'm going to create a `journal-cms--revisions` instance and hand it over for testing.

giorgiosironi:
https://revisions--journal-cms.elifesciences.org/ is ready for your tests (currently `master`, with empty database).

nlisgo:
I'm just restored the db and files. I have pushed this branch to `elifesciences` so that I can deploy it easily to `journal-cms--revisions`.

nlisgo:
The test on `journal-cms--revisions` went as expected but I was surprised that the total number of revisions reported was so different from the one I ran locally a couple of days ago from a db dump taking on the 16th. The revision count was `1703718` up from `1680444`. This indicates that it is still growing and not by a small amount. That's an increase of `23274`.

My recommendation is that we perform this purge and then take another db dump a week after we clean out the revisions and review again where these records are being created.

nlisgo:
I suspect they are being created when we update the metrics of an article. @Dippers your testing has focused on the revisions that are created when we add or modify content in the UI but the article updates are mostly happening programmatically when a queue item arrives and notifies us of a change in the article. I can help you to setup the same circumstances locally without the need for us to implement an SQS queue. We have `go-aws` installed on `drupal-vm`.

Dippers:
I tested the article metrics fetch using the drush commands and even when there was new data (pageview count) to be stored the paragraphs revisions did not increment.

Dippers:
In fact updating articles through the UI does create a new paragraph revisions. This seems to be forced by the entity revisions module even if the node is not being revisioned.

PR #272, opened 2018-02-15
Drush switch msa

nlisgo:


src/modules/jcms_migrate/migration_assets/subjects.json
```diff
@@ -97,6 +97,13 @@
     "image": "subjects/elife-sciences-biophysics-and-structural-biology-illustration.jpg",
     "alt": "Biophysics and Structural Biology"
   },
+  {
+    "id": "biochemistry-chemical-biology",
```
nlisgo: I would prefer if we use migrate to create the subject in production. At least until I refactor how we ensure they are there on a clean install.

giorgiosironi: Preferring that to creating it manually from the UI? I second this opinion.

nlisgo:
Is there any reason that we can't create the new MSA on production now? That should be the only thing holding this up. We need this on `continuumtest--journal-cms` to verify the switch workflow.

src/modules/jcms_migrate/migration_assets/subjects.json
nlisgo: > Preferring that to creating it manually from the UI?

Yes

giorgiosironi:
No idea, would it show in some list as an empty subject? @sianroderick?

Maybe https://github.com/elifesciences/builder-private/pull/136 holds this up as Journal will hide it?

thewilkybarkid:
Yep on both counts (deploying config change now).

src/modules/jcms_migrate/migration_assets/subjects.json
```diff
@@ -97,6 +97,13 @@
...
+    "name": "Biochemistry and Chemical Biology",
+    "impact_statement": "eLife publishes research spanning chemical biology and enzymology to metabolism and membrane proteins. Decisions are made by <a href=\"https://elifesciences.org/about/people/biochemistry\">journal editors</a> who are active researchers in biochemistry.",
+    "image": "subjects/elife-sciences-biochemistry-illustration.jpg",
+    "alt": "Biochemistry and Chemical Biology"
+  },
```
giorgiosironi: shouldn't this show 2 subjects being created?

nlisgo: > shouldn't this show 2 subjects being created?

I'm not sure what you mean, sorry.

nlisgo: Have just added `Immunology and Inflammation`

giorgiosironi: That was it :+1:

PR #273, opened 2018-02-16
Publication date can be scheduled

nlisgo:
Prior to this the publication date has been the date that content was created. This is fine if the content is immediately published but in some instances you want to prepare content prior to publication.

This is solved by enabling the `scheduler` module and to enable a schedule date to be set for relevant content types. You can set a date in the future and once this date has passed the content will be published in a subsequent `cron` run and the published date stored.

Additionally, the date that content switches from unpublished to published is recorded and used as the created date.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
     "drupal/confi": "^1.4",
     "webmozart/path-util": "^2.3",
     "sebastian/version": "~1.0.6",
-    "drupal/publication_date": "1.x-dev"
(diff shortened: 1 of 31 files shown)


nlisgo:
I'm going to take on the work shortly to allow unpublished content to be available to a preview instance of `journal`. I will introduce a test there to cover this work.

nlisgo:
Most changes here are configuration.

PR #277, opened 2018-02-26
Add subject aims and scope field

nlisgo:


src/modules/jcms_rest/src/Plugin/rest/resource/SubjectListRestResource.php
```diff
@@ -84,6 +84,11 @@ public function getItem(EntityInterface $term) {
         unset($item['impactStatement']);
       }
     }
+
+    if ($aims = $this->processFieldContent($term->get('field_aims_and_scope'))) {
```
giorgiosironi: One thing I didn't know is the `/subjects` list contains full items, not just snippets, which explains why this field appears here.

I would have expected this to appear in `SubjectItemRestResource` as well however (or in whatever answers `/subjects/developmental-biology-stem-cells`)?

nlisgo: I have added this now to the subject item

PR #282, opened 2018-03-06
Simplify the aims and scope field

nlisgo:
The aims and scope field is currently using paragraphs. We want to support a single paragraph for now.

This is basically hand holding for a configuration change. Because I want to use the field name `field_aims_and_scope` but adjust the field, I must remove the field and then import the new configuration. The update hook takes care of retrieving the values, removing the field, importing the new config and storing the values again.


giorgiosironi:
Went through it, but I'm not much of an help here.

PR #284, opened 2018-03-09
Allow array or string as sort by value

nlisgo:


src/modules/jcms_rest/src/Plugin/rest/resource/AbstractRestResourceBase.php
```diff
@@ -351,8 +351,13 @@ protected function filterSubjects(QueryInterface &$query) {
...
    * Apply filter for page, per-page and order.
+   *
+   * @param \Drupal\Core\Entity\Query\QueryInterface $query
+   *   Query to filter.
+   * @param string|array|null $sort_by
+   *   Sort by value(s).
    */
-  protected function filterPageAndOrder(QueryInterface &$query, string $sort_by = NULL) {
+  protected function filterPageAndOrder(QueryInterface &$query, $sort_by = NULL) {
     $sort_bys = (array) $this->setSortBy($sort_by);
```
giorgiosironi: setSortBy seems to accept this

PR #287, opened 2018-03-13
Complete list of smoke tests

giorgiosironi:
Looking at PathMediaTypeMapper it seems we are missing some API checks after every deploy.

### DIFF
--- a/smoke_tests.sh
+++ b/smoke_tests.sh
@@ -21,8 +21,8 @@ echo "APIs"
 [ $(curl --write-out %{http_code} --silent --output /dev/null "${hostname}/job-adverts") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null "${hostname}/labs-posts") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null "${hostname}/people") == 200 ]
-[ $(curl --write-out %{http_code} --silent --output /dev/null "${hostname}/press-packages") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null "${hostname}/podcast-episodes") == 200 ]
+[ $(curl --write-out %{http_code} --silent --output /dev/null "${hostname}/press-packages") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null "${hostname}/subjects") == 200 ]
 
 echo "Redis"
--- a/smoke_tests.sh
+++ b/smoke_tests.sh
@@ -11,13 +11,17 @@ echo "Homepage"
 [ $(curl --write-out %{http_code} --silent --output /dev/null "$hostname") == 200 ]
 
 echo "APIs"
+[ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.annual-report-list+json; version=1" "${hostname}/blog-articles") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.blog-article-list+json; version=1" "${hostname}/blog-articles") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.collection-list+json; version=1" "${hostname}/collections") == 200 ]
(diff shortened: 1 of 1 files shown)


smoke_tests.sh
```diff
@@ -11,13 +11,17 @@ echo "Homepage"
 [ $(curl --write-out %{http_code} --silent --output /dev/null "$hostname") == 200 ]
 
 echo "APIs"
+[ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.annual-report-list+json; version=1" "${hostname}/blog-articles") == 200 ]
```
nlisgo: Incorrect path: should be /annual-reports

smoke_tests.sh
```diff
@@ -11,13 +11,17 @@ echo "Homepage"
...
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.blog-article-list+json; version=1" "${hostname}/blog-articles") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.collection-list+json; version=1" "${hostname}/collections") == 200 ]
+[ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.community-list+json; version=1" "${hostname}/covers") == 200 ]
```
nlisgo: Incorrect path

smoke_tests.sh
```diff
@@ -11,13 +11,17 @@ echo "Homepage"
...
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.cover-list+json; version=1" "${hostname}/covers") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.event-list+json; version=1" "${hostname}/events") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.interview-list+json; version=1" "${hostname}/interviews") == 200 ]
+[ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.job-advert-list+json; version=1" "${hostname}/interviews") == 200 ]
```
nlisgo: Incorrect path

smoke_tests.sh
```diff
@@ -11,13 +11,17 @@ echo "Homepage"
...
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.labs-post-list+json; version=1" "${hostname}/labs-posts") == 200 ]
 [ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.person-list+json; version=1" "${hostname}/people") == 200 ]
+[ $(curl --write-out %{http_code} --silent --output /dev/null -H "Accept: application/vnd.elife.press-package-list+json; version=1" "${hostname}/people") == 200 ]
```
nlisgo: Incorrect path

giorgiosironi: Copy&paste :)

giorgiosironi:
Weirdly the wrong smoke tests passed:
```
17:25:54 [34.224.22.179] out: ++ curl --write-out '%{http_code}' --silent --output /dev/null -H 'Accept: application/vnd.elife.annual-report-list+json; version=1' ci--journal-cms.elifesciences.org/blog-articles
17:25:54 [34.224.22.179] out: + '[' 200 == 200 ']'
```
Maybe I should just remove the `Accept` headers, which also need `version` to be updated, and just target the paths? That would also make for a clean `for` loop rather than rewriting all the curl commands.

giorgiosironi:
Removing all `Accept` for simplicity, won't need to track API versions. Dependency on particular `Accept` headers should not be tested with smoke tests but with PHPUnit.

giorgiosironi:
Let's :shipit:

giorgiosironi:
However, `continuumtest` is currently being switched to a test branch so can't merge.

nlisgo:
Merged in latest changes in `develop`

PR #288, opened 2018-03-20
ELPP-3513 Add journal links

nlisgo:
In order for the links to appear we will need to amend the `local.settings.php` for an instance of `journal-cms`. See: https://github.com/elifesciences/journal-cms/pull/288/files#diff-801dc01e1e6ce986ec8619e987d6ce95


src/modules/jcms_admin/src/Plugin/views/field/ContentLink.php
```diff
@@ -0,0 +1,145 @@
...
+      return sprintf('/%s', implode($seperator, array_filter([$stem, $id, $slug])));
+    }
+    elseif ($content->bundle() === 'article') {
+      return sprintf('/%s/%s', 'articles', $content->getTitle());
+    }
+    elseif ($content->bundle() === 'highlight_list') {
+      return sprintf('/%s', $content->getTitle());
+    }
+    elseif ($content->bundle() === 'job_advert') {
+      return sprintf('/%s/%s/%s', 'jobs', substr($content->uuid(), -8), \Drupal::service('slugify')->slugify($content->getTitle()));
```
giorgiosironi: `substr($content->uuid(), -8)` is duplication that can be extracted as a private method

src/modules/jcms_admin/src/Plugin/views/field/ContentLink.php
```diff
@@ -0,0 +1,145 @@
...
+    }
+    elseif ($content->bundle() === 'podcast_chapter') {
+      if ($episode = $this->getChapterEpisodeNumber($content)) {
+        return sprintf('/%s%d#%d', 'podcast/episode', $episode, (int) $content->get('field_podcast_chapter_time')->getString());
+      }
+    }
+    elseif ($content->bundle() === 'press_package') {
+      return sprintf('/%s/%s/%s', 'for-the-press', substr($content->uuid(), -8), \Drupal::service('slugify')->slugify($content->getTitle()));
```
giorgiosironi: there's obviously a coupling to journal, but this being public URLs should be very stable or at least get redirected if they change

src/modules/jcms_admin/src/Plugin/views/field/ContentLink.php
```diff
@@ -0,0 +1,145 @@
...
+          break;
+
+        case 'labs_experiment':
+          $stem = 'labs';
+          $slug = \Drupal::service('slugify')->slugify($content->getTitle());
+          break;
+
+        case 'podcast_episode':
+          $stem = 'podcast/episode';
+          $seperator = '';
```
giorgiosironi: typo?

src/modules/jcms_admin/src/Plugin/views/field/ContentLink.php
```diff
@@ -0,0 +1,145 @@
...
+      $slug = NULL;
+      $seperator = '/';
+      switch ($content->bundle()) {
+        case 'annual_report':
+          $id = NULL;
+          break;
+
+        case 'blog_article':
+          $stem = 'inside-elife';
+          $slug = \Drupal::service('slugify')->slugify($content->getTitle());
```
giorgiosironi: not sure if the slugify algorithm here will always be 100% identical to what `journal` does. However, `journal` will accept also incorrect slugs redirecting them to the canonical URL. So this is robust :+1:

config/drupal-vm.services.yml
```diff
@@ -2,6 +2,9 @@ parameters:
   monolog.channel_handlers:
     default: ['default_handler', 'error_handler']
   monolog.processors: ['process_id', 'message_placeholder', 'current_user', 'request_uri', 'ip', 'referer']
+  journal:
+    path: https://elifesciences.org
```
giorgiosironi: URLs usually have been put into https://github.com/elifesciences/journal-cms-formula/blob/master/salt/journal-cms/config/srv-journal-config-local.settings.php rather than https://github.com/elifesciences/journal-cms-formula/blob/master/salt/journal-cms/config/srv-journal-config-local.services.yml
Can the same be done here?

src/modules/jcms_admin/src/Plugin/views/field/ContentLink.php
thewilkybarkid: See also ELPP-3070.

src/modules/jcms_admin/src/Plugin/views/field/ContentLink.php
thewilkybarkid: Could also not worry about the slugs.

nlisgo: I will check if dropping the slugs is acceptable.

PR #293, opened 2018-04-04
ELPP-3576 Drupal 8.5.1

nlisgo:
Used: `composer update drupal/core symfony/config --with-dependencies`

See: https://github.com/composer/composer/issues/7234

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -122,7 +122,8 @@
         "View output is not used for autocomplete display (2174633)": "https://www.drupal.org/files/issues/2174633-198.patch",
         "Use new Transliteration functionality in core for file names (2492171)": "https://www.drupal.org/files/issues/use_new_transliteration-2492171-51.patch",
         "Move permission \"view any unpublished content\" from content moderation to core (273595)": "https://www.drupal.org/files/issues/move_permission_view-273595-129.patch",
-        "NegotiationMiddleware calls $request->setRequestFormat('html') when there is no _format request parameter, but shouldn't (2854543)": "https://www.drupal.org/files/issues/2018-04-03/2854543-57.patch"
+        "NegotiationMiddleware calls $request->setRequestFormat('html') when there is no _format request parameter, but shouldn't (2854543)": "https://www.drupal.org/files/issues/2018-04-03/2854543-57.patch",
+        "Temporary fix for \\Drupal\\Core\\Field\\FieldItemList::generateSampleItems (2905527)": "src/patches/2905527-temporary-fix-generate-sample-items.patch"
       },
       "drupal/entity_reference_revisions": {
         "Allow entities to be generated if none exist (2910326)": "https://www.drupal.org/files/issues/entity_reference_revisions-stop_checking_for_entities_before_generating_sample-2910326-2.patch"
--- a/composer.json
+++ b/composer.json
@@ -77,6 +77,9 @@
   "conflict": {
     "drupal/drupal": "*"
   },
+  "config": {
+    "platform": {"php": "7.0.17"}
+  },
   "autoload": {
     "classmap": [
       "scripts/composer/ScriptHandler.php"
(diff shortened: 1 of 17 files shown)


giorgiosironi:
The tests seem to fail due to the underlying database, but it's consistent in Alfred and Travis CI so it should be easy to reproduce.

composer.json
```diff
@@ -118,7 +121,9 @@
       "drupal/core": {
         "View output is not used for autocomplete display (2174633)": "https://www.drupal.org/files/issues/2174633-198.patch",
         "Use new Transliteration functionality in core for file names (2492171)": "https://www.drupal.org/files/issues/use_new_transliteration-2492171-51.patch",
-        "Move permission \"view any unpublished content\" from content moderation to core (273595)": "https://www.drupal.org/files/issues/move_permission_view-273595-129.patch"
+        "Move permission \"view any unpublished content\" from content moderation to core (273595)": "https://www.drupal.org/files/issues/move_permission_view-273595-129.patch",
+        "NegotiationMiddleware calls $request->setRequestFormat('html') when there is no _format request parameter, but shouldn't (2854543)": "https://www.drupal.org/files/issues/2018-04-03/2854543-57.patch",
+        "Temporary fix for \\Drupal\\Core\\Field\\FieldItemList::generateSampleItems (2905527)": "src/patches/2905527-temporary-fix-generate-sample-items.patch"
```
nlisgo: Since the following fix, you need to introduce field type plugins in a different way: https://www.drupal.org/node/2905527

I will create a separate ticket to change how we generate sample values for specific fields.

composer.json
```diff
@@ -77,6 +77,9 @@
   "conflict": {
     "drupal/drupal": "*"
   },
+  "config": {
+    "platform": {"php": "7.0.17"}
```
nlisgo: This should ensure that no dependencies come down which will be incompatible with the PHP version on our production environment.

src/modules/jcms_rest/src/EventSubscriber/ExceptionJcmsJsonSubscriber.php
```diff
@@ -19,15 +19,6 @@ protected function getHandledFormats() {
     return ['jcms_json'];
   }
 
-  /**
-   * {@inheritdoc}
-   */
-  protected static function getPriority() {
-    // This will fire after the most common HTML handler, since HTML requests
-    // are still more common than JSON requests.
-    return -75;
```
nlisgo: Since updating to `8.5.1` this was not being prioritized highly enough. It fixed the issue if I removed it and relied upon the default priority.

src/modules/jcms_rest/src/Routing/RouteSubscriber.php
```diff
@@ -18,7 +18,11 @@ class RouteSubscriber extends RouteSubscriberBase {
    * {@inheritdoc}
    */
   public static function getSubscribedEvents() {
-    return [RoutingEvents::ALTER => ['onAlterRoutes', -9999]];
+    $events = parent::getSubscribedEvents();
+
+    $events[RoutingEvents::ALTER] = ['onAlterRoutes', -9999];
+
+    return $events;
```
nlisgo: This was not needed but it seemed to be the incorrect approach to disregard other subscribed events.

src/modules/jcms_rest/src/Routing/RouteSubscriber.php
```diff
@@ -29,7 +33,7 @@ protected function alterRoutes(RouteCollection $collection) {
       // If the route matches a JCMS rest endpoint.
       if (preg_match("/^rest.[a-z_]+.[A-Z]+.jcms_json$/", $name)) {
         // Remove the request_format_route_filter from the route _filters.
-        $route_filters = $route->getOption('_route_filters');
+        $route_filters = $route->getOption('_route_filters') ?: [];
```
nlisgo: This was causing a warning on a clean install of the site.

composer.json
giorgiosironi: agreed. Do you need the patch version or can it just be `7.0`?

PR #303, opened 2018-05-01
ELPP-3475 Remove MSA genes-chromosomes

nlisgo:
Also upgrade recommeded contrib modules


thewilkybarkid:
We need to add in Journal redirects first.

nlisgo:
> We need to add in Journal redirects first.

https://elifesciences.org/subjects/genes-chromosomes is currently redirecting to https://elifesciences.org/subjects/chromosomes-gene-expression

thewilkybarkid:
That's the rewrite (which we also need to remove).

PR #304, opened 2018-05-03
ELPP-2762 Expose competingInterests

nlisgo:
~~I will do a followup commit to resave all the content that has a competing interest so that the reindex is triggered.~~ DONE

### DIFF
--- a/src/modules/jcms_migrate/jcms_migrate.install
+++ b/src/modules/jcms_migrate/jcms_migrate.install
@@ -1062,3 +1062,43 @@ function jcms_migrate_update_8124() {
     $migrate_id_map->delete([$id]);
   }
 }
+
+/**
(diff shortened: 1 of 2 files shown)


src/modules/jcms_migrate/jcms_migrate.install
```diff
@@ -1062,3 +1062,43 @@ function jcms_migrate_update_8124() {
     $migrate_id_map->delete([$id]);
   }
 }
+
+/**
+ * Resave all person nodes with a competing interest.
+ */
+function jcms_migrate_update_8125(&$sandbox) {
+  if (!isset($sandbox['progress'])) {
```
nlisgo: This is probably overkill because so few nodes are affected. I will refactor.

PR #305, opened 2018-05-08
ELPP-2762 Allow MSA switch script to work for orphaned MSA

nlisgo:
Continuumtest has some orphaned MSA. The `../vendor/bin/drush msa-switch` script must not have been run on this instance before an MSA was removed. 

This script allows us to target the Drupal term id as the `from` value of the `msa-switch` command.

For example:

```
cd /srv/journal-cms/web
../vendor/bin/drush msa-switch 7 chromosomes-gene-expression
```

### DIFF
--- a/scripts/sanitized-db-dump.sh
+++ b/scripts/sanitized-db-dump.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-DEFAULT_STRUCTURED_TABLES_LIST=" --structure-tables-list='user__roles,user__user_picture,users,users_data,users_field_data'"
+DEFAULT_STRUCTURED_TABLES_LIST=" --structure-tables-list=user__roles,user__user_picture,users,users_data,users_field_data"
 
 if [[ ${*} =~ \-\-structure\-tables\-list ]];
 then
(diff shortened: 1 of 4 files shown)


nlisgo:
> I guess this is run manually so the user can force an id

Yes it is. We should discuss the impact on continuumtest on future MSA changes.

giorgiosironi:
I don't know the whole process, but I would have expected that if we migrate some data in `prod`, the same migration is added to the repository so that it runs on all environments. Assuming:

- that migration would be idempotent on `prod` or it would be recognized as "already run" and not run again there
- `continuumtest` is the only environment that persists data other than `prod` (`end2end` and `ci` are routinely cleaned)

Of course in a perfect world we would commit a migration, and it would go through all environments automatically: `end2end`->`continuumtest`->`prod`.

nlisgo:
The trigger for switching the MSA of content unique to `journal-cms` is a drush command that is run manually. It has been run on `prod` but not on any other environment routinely. After the switch happens and all of the article content in `journal-cms` has been recategorised then we remove the MSA entirely.

Where the MSA switch is simply an allocation of all content assign to MSA `A` to MSA `B` then it is feasible to trigger that programmatically but in other instances we may need staff to manually adjust the content if the reassignment requires some thought.

To mitigate this we could ensure that when we remove the MSA entirely from `journal-cms` that we agree to assign any content that is still categorised under the outgoing MSA to a specific remaining MSA. This could be done programmatically.

nlisgo:
I have added the revert of the `competingInterests` work to this PR so that we can cleanup `continuumtest` ahead of that release.

PR #307, opened 2018-05-09
ELPP-3591 Reduce the minimum image size for labs posts

nlisgo:


src/modules/jcms_migrate/jcms_migrate.install
```diff
@@ -1081,3 +1081,20 @@ function jcms_migrate_update_8125() {
...
+  $nids = $query->execute();
+  if ($nids) {
+    $nodes = Node::loadMultiple($nids);
+    if (!empty($nodes)) {
+      foreach ($nodes as $node) {
+        $node->save();
+      }
+    }
+  }
+}
```
nlisgo: The resave isn't strictly necessary because it will not impact the display of labs posts content but the API response will be altered because we no longer present a banner image as it is not being used.

PR #310, opened 2018-05-21
ELPP-3593 Remove genomics-evolutionary-biology msa

nlisgo:
I will submit PR's to `journal-formula` and `builder-private` to setup the redirects and remove the rewrite that we have in place.

### DIFF
--- a/src/modules/jcms_migrate/jcms_migrate.install
+++ b/src/modules/jcms_migrate/jcms_migrate.install
@@ -1121,3 +1121,29 @@ function jcms_migrate_update_8128() {
   $executable = new MigrateExecutable($plugins['jcms_subjects_json'], new MigrateMessage());
   $executable->import();
 }
+
+/**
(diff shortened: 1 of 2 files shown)


src/modules/jcms_migrate/jcms_migrate.install
```diff
@@ -1121,3 +1121,29 @@ function jcms_migrate_update_8128() {
...
   $executable->import();
 }
+
+/**
+ * Remove genomics-evolutionary-biology msa.
+ */
+function jcms_migrate_update_8129() {
+  // Assign any remaining content to genetics-genomics, there shouldn't be any
+  // in production.
+  drush_invoke('msa-switch', ['genomics-evolutionary-biology', 'genetics-genomics']);
```
nlisgo: Running this command will ensure that content in `continuumtest` has no reference to the outgoing MSA.

PR #312, opened 2018-06-06
Update drupal/core to null

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-journal-cms-update-drupal/19/display/redirect which resulted in this pull request.


giorgiosironi:
Was a wrong replay.

PR #318, opened 2018-06-07
Update drupal/core to 8.5.4

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-journal-cms-update-drupal/29/display/redirect which resulted in this pull request.


giorgiosironi:
Ah, this was created after the `8.5.3`, and we merged that. So now it shows a conflict. Better to close this and let the pipeline run again, to start from the `8.5.3` develop.

PR #324, opened 2018-07-04
Fix IIIF identifiers

thewilkybarkid:


thewilkybarkid:
No failing tests? 😒

thewilkybarkid:
No failing tests? 😒

src/modules/jcms_rest/src/JCMSImageUriTrait.php
```diff
@@ -127,4 +131,21 @@ protected function getImageSizes($size_types = ['banner', 'thumbnail']) : array
     return $sizes;
   }
 
+  /**
```
giorgiosironi: this trait is used in a Drupal hook and in an abstract class, so two options to test it:
- test class that also defines a class including the trait
- https://github.com/elifesciences/journal-cms/blob/361cabe4f662f739d69be56a6e6b99ceaf116e77/src/modules/jcms_rest/tests/src/Functional/RamlSchemaValidationTest.php or https://github.com/elifesciences/journal-cms/blob/361cabe4f662f739d69be56a6e6b99ceaf116e77/src/modules/jcms_rest/tests/src/Functional/RecursiveEndpointValidatorTest.php

The first involves dependencies on `FieldItemListInterface` and several `Drupal::` calls so does not seem feasible

thewilkybarkid: The second doesn't actually check that the computed URI is correct, just that it matches the schema.

giorgiosironi: The current API tests do not look to delve into the response IIIF-related fields, but they're the best best for extension as they treat `journal-cms` as a black box, saving on setup complexity

nlisgo: I will add a test on a separate branch that we can see fail and then add in with this fix.

nlisgo: Here is a test: https://github.com/elifesciences/journal-cms/pull/326

It doesn't cover everything in the trait but enough for this issue.

thewilkybarkid:
Should be able to :shipit:.

PR #326, opened 2018-07-05
Test for IIIF identifiers

nlisgo:
This test should fail and the fix is here: https://github.com/elifesciences/journal-cms/pull/324

### DIFF
--- a/src/modules/jcms_rest/tests/src/Unit/JCMSImageUriTraitTest.php
+++ b/src/modules/jcms_rest/tests/src/Unit/JCMSImageUriTraitTest.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Drupal\Tests\jcms_rest\Unit;
+
+use Drupal\Core\Site\Settings;
+use Drupal\jcms_rest\JCMSImageUriTrait;
+use Drupal\Tests\UnitTestCase;
+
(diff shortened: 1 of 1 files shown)


giorgiosironi:
The failures look understandable. Probably to be merge into the other branch rather than `develop`.

src/modules/jcms_rest/tests/src/Unit/JCMSImageUriTraitTest.php
```diff
@@ -0,0 +1,58 @@
...
+use Drupal\Tests\UnitTestCase;
+
+/**
+ * Tests for methods in JCMSImageUriTrait.
+ *
+ * @package Drupal\Tests\jcms_rest\Unit
+ */
+class JCMSImageUriTraitTest extends UnitTestCase {
+
+  use JCMSImageUriTrait;
```
thewilkybarkid: I'd say that mixing production code with tests shows that `JCMSImageUriTrait` needs to be refactored, but this is better than nothing.

thewilkybarkid:
Changed the base and re-running.

thewilkybarkid:
`travis-ci/pr` should be the only green build as it tests the PR not the branch.

PR #328, opened 2018-07-09
[WIP] Update people to list by multiple types

thewilkybarkid:


thewilkybarkid:
Don't think there's any test coverage here.

src/modules/jcms_rest/src/Plugin/rest/resource/PersonListRestResource.php
```diff
@@ -102,18 +101,10 @@ protected function filterSubjects(QueryInterface &$query) {
...
    * Apply filter for type by amending query.
    */
-  protected function filterType(QueryInterface &$query) {
-    $type = $this->getRequestOption('type');
-    if (!is_null($type)) {
-      $entityManager = \Drupal::service('entity_field.manager');
-      $fields = $entityManager->getFieldStorageDefinitions('node', 'person');
-      $options = options_allowed_values($fields['field_person_type']);
-      if (!in_array($type, array_keys($options))) {
-        throw new JCMSBadRequestHttpException(t('Invalid type'));
```
nlisgo: We're losing the ability to inform the client that the type is not recognised but this isn't strictly required. We may want to distinguish between a `type` that produces no results and a `type` that isn't recognised. Don't feel strongly about it. This can go.

src/modules/jcms_rest/src/Plugin/rest/resource/PersonListRestResource.php
```diff
@@ -102,18 +101,10 @@ protected function filterSubjects(QueryInterface &$query) {
...
-      }
-      else {
-        $query->condition('field_person_type.value', $type);
-      }
+  protected function filterTypes(QueryInterface &$query) {
+    $types = (array) $this->getRequestOption('type');
+    if (!empty($types)) {
+      $query->condition('field_person_type.value', $types, 'IN');
```
nlisgo: Just want to confirm that this is supposed to be an `OR` query. I will just run this locally to confirm that if a person matches on multiple types that they don't appear twice in the listing. I agree we need better test coverage on this.

nlisgo: `type` has a cardinality of 1 so there is no risk of a person appearing twice. This is an `OR` query. Happy to ship this.

PR #334, opened 2018-07-26
Bump versions of annual-report-list and annual-report

nlisgo:
Improves: https://github.com/elifesciences/jira-import/issues/4192

### DIFF
--- a/src/modules/jcms_rest/src/Plugin/rest/resource/AnnualReportItemRestResource.php
+++ b/src/modules/jcms_rest/src/Plugin/rest/resource/AnnualReportItemRestResource.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\jcms_rest\Plugin\rest\resource;
 
+use Drupal\jcms_rest\Exception\JCMSNotAcceptableHttpException;
 use Drupal\node\Entity\Node;
 use Drupal\jcms_rest\Exception\JCMSNotFoundHttpException;
 use Drupal\jcms_rest\Response\JCMSRestResponse;
@@ -20,6 +21,7 @@ use Symfony\Component\HttpFoundation\Response;
  * )
  */
 class AnnualReportItemRestResource extends AbstractRestResourceBase {
+  protected $latestVersion = 2;
 
   /**
    * Responds to GET requests.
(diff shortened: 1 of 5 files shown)


nlisgo:
In order for this to pass we need to merge in: https://github.com/elifesciences/api-raml/pull/224

src/modules/jcms_rest/src/Plugin/rest/resource/AnnualReportItemRestResource.php
```diff
@@ -61,8 +63,15 @@ public function get(int $year) : JCMSRestResponse {
...
-      // Image is required.
-      $response['image'] = $this->processFieldImage($node->get('field_image'), FALSE, 'thumbnail', TRUE);
+      // Image is required, for version 1.
+      if ($this->acceptVersion < 2) {
+        if ($image = $this->processFieldImage($node->get('field_image'), FALSE, 'thumbnail', TRUE)) {
+          $response['image'] = $image;
+        }
+        else {
+          throw new JCMSNotAcceptableHttpException('This annual report requires version 2+.');
+        }
```
giorgiosironi: This means no new annual reports should be added until version 1 is not used by Journal anymore, but from the name `annual` I guess there's time for that

src/modules/jcms_rest/src/Plugin/rest/resource/AnnualReportListRestResource.php
```diff
@@ -87,8 +87,15 @@ public function getItem(EntityInterface $node) : array {
...
 
-    // Image is required.
-    $item['image'] = $this->processFieldImage($node->get('field_image'), FALSE, 'thumbnail', TRUE);
+    // Image is required, for version 1.
+    if ($this->acceptVersion < 2) {
+      if ($image = $this->processFieldImage($node->get('field_image'), FALSE, 'thumbnail', TRUE)) {
+        $item['image'] = $image;
+      }
+      else {
+        throw new JCMSNotAcceptableHttpException('This annual report list requires version 2+.');
```
giorgiosironi: Since this is a listing, may be `At least one of the annual report in the list requires version 2+.`?

src/modules/jcms_rest/src/Plugin/rest/resource/AnnualReportItemRestResource.php
nlisgo: Since image is still a required field there is no risk of version 1 not being support. I will drop images from `journal-cms` after we have moved `journal` to version 2.

PR #336, opened 2018-08-06
Update drupal/core to 8.5.6

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-journal-cms-update-drupal/149/display/redirect which resulted in this pull request.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -6980,25 +6980,28 @@
         },
         {
             "name": "symfony/polyfill-ctype",
-            "version": "v1.8.0",
+            "version": "v1.9.0",
             "source": {
                 "type": "git",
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Tricky: Drupal is already on 8.5.6 but since its update involves its dependencies (many `symfony/*` packages), those have been updates to 1.9.0 hence there was something updated for a PR to capture.

giorgiosironi:
So still think that:
- this should be integrated
- we should drop Travis CI if it's not working

nlisgo:
> we should drop Travis CI if it's not working

See: #338

PR #338, opened 2018-08-13
Drop travis support

nlisgo:


giorgiosironi:
Build error doesn't seem related (could have been if a script used by the build has been removed?)

PR #341, opened 2018-08-31
Deploy inline editor and migrate content

nlisgo:


composer.json
```diff
@@ -74,14 +299,38 @@
...
+    "ckeditor/uploadimage": "^4.9",
+    "ckeditor/uploadwidget": "^4.9",
+    "ckeditor/widget": "^4.9",
+    "drupal/hide_revision_field": "^1.1",
+    "ckeditor/link": "^4.9",
+    "ckeditor/fakeobjects": "^4.9",
+    "drupal/field_group": "~3.0-beta1",
     "elife/api": "dev-master",
-    "elife/api-validator": "dev-master",
+    "elife/api-validator": "^1.0@dev",
```
nlisgo: We now perform schema checks prior to allowing an editor to publish content.

sync/core.extension.yml
```diff
@@ -9,6 +9,7 @@ module:
   ckeditor: 0
   config: 0
   config_devel: 0
+  config_import: 0
```
nlisgo: I don't believe this is needed

PR #356, opened 2018-10-22
Update drupal/core to 8.6.2

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-journal-cms-update-drupal/303/display/redirect which resulted in this pull request.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -5791,16 +5791,16 @@
         },
         {
             "name": "masterminds/html5",
-            "version": "2.3.0",
+            "version": "2.3.1",
             "source": {
                 "type": "git",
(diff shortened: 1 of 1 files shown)


giorgiosironi:
This was already done in https://github.com/elifesciences/journal-cms/pull/355 but the [command we run to upgrade](https://github.com/elifesciences/journal-cms/blob/develop/Jenkinsfile.update-drupal#L7) has some dependencies to be updated in lockstep with Drupal. In this case only one of those dependencies have changed.

giorgiosironi:
(opinion: this should be good to go)

PR #369, opened 2018-11-14
Spaces in image names

nlisgo:
Improves: https://github.com/elifesciences/issues/issues/4465


src/modules/jcms_ckeditor/js/jcms_ckeditor.js
```diff
@@ -277,7 +277,7 @@
                     type: "file--image",
                     attributes: {
                       data: image[1],
-                      uri: 'public://' + settings.imageFileDirectory + '/' + e.data.fileLoader.fileName
+                      uri: 'public://' + settings.imageFileDirectory + '/' + e.data.fileLoader.fileName.replace(/\s+/g, '-').toLowerCase()
```
nlisgo: @davidcmoulton any objection to this approach. The aim is to clear the filename of spaces and to ensure that the extension is in lowercase.

PR #381, opened 2018-12-11
Download only the more recent backup

giorgiosironi:
Sorts in reverse order according to the first column (modification date) and selects the first database dump or files archive:
```
$ aws s3 ls elife-app-backups/journal-cms/201812/201812 | sort -r | head -n 1
2018-12-10 23:15:45 4517142383 20181210_prod--journal-cms.elifesciences.org_231544-archive-b47198f6.tar.gz
```


nlisgo:
@giorgiosironi, apologies for the delay in review.

PR #401, opened 2019-03-20
Update drupal/core to 8.6.13

nlisgo:
The update drupal pipeline didn't produce a PR because one of the patches did not apply: https://alfred.elifesciences.org/blue/organizations/jenkins/dependencies%2Fdependencies-journal-cms-update-drupal/detail/dependencies-journal-cms-update-drupal/582/pipeline

The patch that did not apply was:

```
"Use new Transliteration functionality in core for file names (2492171)": "https://www.drupal.org/files/issues/2019-01-29/2492171-3-151.patch",
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -384,7 +384,7 @@
     "patches": {
       "drupal/core": {
         "View output is not used for autocomplete display (2174633)": "https://www.drupal.org/files/issues/2174633-198.patch",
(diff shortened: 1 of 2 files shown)


nlisgo:
@giorgiosironi it is great that a patch failing to apply will block deployment. When it fails, how disruptive is it?

giorgiosironi:
Normally the dependency update tries to just produce a PR, and that PR is tested independently. However it does have to successfully run `composer update ...` to produce a new `composer.lock`, so it can't fail later on.
It's not disruptive to the environment that is being used, but someone should get notified of the failure?

PR #408, opened 2019-04-16
Fix tables that contain non-ASCII characters

thewilkybarkid:
Slightly alarmed by this code, but seems to fix https://github.com/elifesciences/issues/issues/4878.

### DIFF
--- a/src/modules/jcms_admin/src/HtmlMarkdownSerializer.php
+++ b/src/modules/jcms_admin/src/HtmlMarkdownSerializer.php
@@ -100,7 +100,7 @@ final class HtmlMarkdownSerializer implements NormalizerInterface {
       }
     }
     $preserve = implode('', $lines);
-    return preg_replace_callback('~' . $bc . '(code|table)[^' . $bc . ']*' . $bc . '([^' . $bc . ']*)' . $bc . '/\1' . $bc . '~s', function ($matches) use ($bc, $encode) {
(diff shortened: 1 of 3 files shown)


thewilkybarkid:
As far as I can tell, it's converting HTML like `<table><tr><td>Cell one</td></tr></table>` to `øtableø<tr><td>Cell one</td></tr>ø/tableø` to then base64-encode the contents so they _don't_ go through a Markdown processor.

thewilkybarkid:
The UTF-8 modifier (`u`) also works, but would still fail on content containing `ø`.

src/modules/jcms_admin/src/HtmlMarkdownSerializer.php
```diff
@@ -100,7 +100,7 @@ private function preserveOutput(string $html, array $context = []) : string {
       }
     }
     $preserve = implode('', $lines);
-    return preg_replace_callback('~' . $bc . '(code|table)[^' . $bc . ']*' . $bc . '([^' . $bc . ']*)' . $bc . '/\1' . $bc . '~s', function ($matches) use ($bc, $encode) {
```
giorgiosironi: can't understand why `$bc` was `ø`, but ok

nlisgo: It was just an arbitrary character.

PR #425, opened 2019-07-24
Drupal 8.7.5 and contrib updates

nlisgo:


nlisgo:
Close in favour of: https://github.com/elifesciences/journal-cms/pull/428

PR #449, opened 2019-11-06
Update commonmark library

nlisgo:


nlisgo:
The expected output is different because of this: 

```
https://github.com/thephpleague/commonmark/commit/92ab7cb0a6d3607e35d3fc5f93c868db50c03fdc#diff-cf08af3de68178d608e349719d010d2c
```

PR #489, opened 2020-02-18
Expose subjects in press packages

nlisgo:


src/modules/jcms_migrate/jcms_migrate.install
```diff
@@ -1563,3 +1563,41 @@ function jcms_migrate_update_8150() {
     }
   }
 }
+
+/**
+ * Resave all press_package nodes for re-index.
```
nlisgo: The object of the resave is to trigger re-indexing of content in other services but changes to press-packages aren't generating SNS notifications. Is this a bug or do we not require them because they are not being indexed in the search?

src/modules/jcms_rest/src/Plugin/rest/resource/PressPackageListRestResource.php
```diff
@@ -87,8 +87,8 @@ public function getItem(EntityInterface $node) {
     // Subjects are optional.
     if ($node->get('field_related_content')->count()) {
       $articles = [];
-      foreach ($node->get('field_related_content') as $related) {
-        if ($article = $this->getArticleSnippet($related->get('entity')->getTarget()->getValue())) {
+      foreach ($node->get('field_related_content')->referencedEntities() as $related) {
```
nlisgo: This is just a simplification. It  brings it inline with: https://github.com/elifesciences/journal-cms/blob/develop/src/modules/jcms_rest/src/Plugin/rest/resource/PressPackageItemRestResource.php#L68-L72

PR #497, opened 2020-03-26
Upgrade to drush 9

nlisgo:


nlisgo:
This will need a bit more attention because the commands that we use in ci will need to adjusted.

PR #501, opened 2020-03-27
Promotional collections

nlisgo:


src/modules/jcms_rest/src/Plugin/rest/resource/PromotionalCollectionListRestResource.php
```diff
@@ -0,0 +1,134 @@
...
+ * )
+ */
+class PromotionalCollectionListRestResource extends AbstractRestResourceBase {
+
+  /**
+   * Responds to GET requests.
+   *
+   * Returns a list of bundles for specified entity.
+   *
+   * @todo - elife - nlisgo - Handle version specific requests
```
nlisgo, line 30: These todo's can go because we are handling them

src/modules/jcms_rest/src/Plugin/rest/resource/PromotionalCollectionListRestResource.php
```diff
@@ -0,0 +1,132 @@
...
+      ->condition('changed', \Drupal::time()->getRequestTime(), '<')
+      ->condition('type', 'promotional_collection');
+
+    if (!$this->viewUnpublished()) {
+      $base_query->condition('status', NodeInterface::PUBLISHED);
+    }
+
+    $this->filterSubjects($base_query);
+
+    $containing = \Drupal::request()->query->get('containing', []);
```
nlisgo, line 41: We have `filterContaining` now.

src/modules/jcms_rest/src/Plugin/rest/resource/PromotionalCollectionItemRestResource.php
```diff
@@ -0,0 +1,157 @@
...
+          }
+        }
+
+        // Summary is optional.
+        if ($content = $this->processFieldContent($node->get('field_summary'))) {
+          $response['summary'] = $content;
+        }
+
+        // Promotional collection content is required.
+        $response['content'] = [];
```
nlisgo, line 95: I think it is reasonable to remove some duplicate code here that can be shared with `CollectionItemRestResource`

nlisgo: I will refactor this resource after this is merged in: https://github.com/elifesciences/journal-cms/pull/508

src/modules/jcms_rest/src/Plugin/rest/resource/PromotionalCollectionItemRestResource.php
```diff
@@ -0,0 +1,157 @@
...
+   * Responds to GET requests.
+   *
+   * Returns a list of bundles for specified entity.
+   *
+   * @throws JCMSNotFoundHttpException
+   */
+  public function get(string $id) : JCMSRestResponse {
+    if ($this->checkId($id)) {
+      $query = \Drupal::entityQuery('node')
+        ->condition('changed', \Drupal::time()->getRequestTime(), '<')
```
giorgiosironi, line 34: what's the time about? Something related to multiple versions?

src/modules/jcms_rest/tests/src/Functional/ValidQueryEndpointTest.php
```diff
@@ -26,6 +26,15 @@ public function validQueryEndpointProvider() : array {
         400,
         'Invalid containing parameter',
       ],
+      'containing - promotional-collections' => [
+        '/collections?containing[]=article/1&containing[]=blog-article/2&containing[]=digest/3&containing[]=event/4&containing[]=interview/5',
```
giorgiosironi, line 30: this query doesn't mention `promotional-collections`, only `collections` and 5 allowed simpler content types?

nlisgo: Will fix, the containing parameter will trigger an error if a single filter value is not supported

src/modules/jcms_rest/src/Plugin/rest/resource/PromotionalCollectionItemRestResource.php
nlisgo: These conditions could go on single item queries. I can't honestly remember the significance of this condition for lists but it might protect against any date scheduled for publication in the future. Will remove for all item queries in separate PR and refactor here.

nlisgo: It's gone.

PR #510, opened 2020-04-08
Improve test coverage for highlight lists and covers lists

nlisgo:


sync/field.field.node.cover.field_cover_content.yml
```diff
@@ -8,7 +8,6 @@ dependencies:
     - node.type.blog_article
     - node.type.collection
     - node.type.cover
-    - node.type.event
```
nlisgo, line 11: These tests exposed the bug that events were able to be introduced into covers even though they are not supported by the schema

PR #519, opened 2020-04-27
fixes bug in restore script

lsh-0:
fixes bug in restore script where multiple found files causes buggy command to be generated


lsh-0:
merged, restore test: https://alfred.elifesciences.org/job/process/job/process-journal-cms-restore-continuumtest/110/console

PR #538, opened 2020-06-03
Support new twitter block in inline editor

nlisgo:


Dippers:
Currently the twitter embed only renders correctly when first created. Any changes in the twitter source url and not reflected in the rendered output. This is because the first rendering removes the oembed element and replaces it with the twitter 'blockquote'. When the source url is edited/changed no oembed element is found so the output is not changed. The following patch does not depend on the oembed being present and instead just removes any previous rendering and then creates the new content.

in jcms_ckeditor/ckeditor/plugins/twitterembed/plugin.js replace _setContent with 
```
_setContent: function (url, content ) {
  this.setData('url', url );
  // Remove all nodes
  while (this.element.$.firstChild) {
    this.element.$.removeChild(this.element.$.lastChild);
  }
  // Create new node and set content
  this.element.$.appendChild(document.createElement('p'));
  this.element.$.childNodes[0].outerHTML = content;
}
```
This change can also be made in the googlemap plugin.

nlisgo:
> Currently the twitter embed only renders correctly when first created.

Thanks for this feedback. I will add this in a followup.

PR #540, opened 2020-06-04
Unpublish libero publisher press package

nlisgo:


nlisgo:
As the press packages are not included in search results we do not need to trigger a search re-index after the deletion.

PR #653, opened 2020-09-16
Styling of bulleted lists is inconsistent in Job Advert posts

NuclearRedeye:
Fix the inconsistent styling of lists in job advert posts.

This fixes elifesciences/issues#6074


nlisgo:
I am going to close this, in favour of https://github.com/elifesciences/journal-cms/pull/653

The test doesn't add value because it the code being tested was not being run.

PR #654, opened 2020-09-16
Fix drupal-vm 2.0.10

nlisgo:
See: https://github.com/geerlingguy/ansible-role-php-versions/pull/51

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -9,3 +9,8 @@ ENV['DRUPALVM_DIR'] = "vendor/geerlingguy/drupal-vm"
 
 # Load the real Vagrantfile
 load "#{__dir__}/#{ENV['DRUPALVM_DIR']}/Vagrantfile"
+
+# Lock to v2.0.10 of the geerlingguy/drupal-vm box image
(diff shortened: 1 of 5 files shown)


Vagrantfile
```diff
@@ -9,3 +9,8 @@ ENV['DRUPALVM_DIR'] = "vendor/geerlingguy/drupal-vm"
 
 # Load the real Vagrantfile
 load "#{__dir__}/#{ENV['DRUPALVM_DIR']}/Vagrantfile"
+
+# Lock to v2.0.10 of the geerlingguy/drupal-vm box image
+Vagrant.configure("2") do |config|
+  config.vm.box_version = "2.0.10"
```
nlisgo, line 15: Will verify that this works. Happy for this to get in.

NuclearRedeye (approved):
Tested locally, and this has now resolved the issues with getting the local image up and running. I added a patch to fix the versions so that the goal posts don't move again in the future without us knowing.

NuclearRedeye:
@nlisgo You might want to review the patch I added to make sure that you're happy.

PR #666, opened 2020-09-25
Ensure jobs are using the same code to parse html as other content types

nlisgo:


src/modules/jcms_rest/tests/src/Functional/RecursiveEndpointValidatorTest.php
```diff
@@ -56,90 +56,88 @@ public function setUp() {
    */
   public function dataProvider() : array {
     return [
-      [
+      '/subjects' => [
```
nlisgo, line 59: Changes to this test fixture were made after repeatedly running it against content imported from a recent database dump. This ensures that better feedback is available for failing tests.

src/modules/jcms_migrate/jcms_migrate.install
```diff
@@ -1699,3 +1699,48 @@ function jcms_migrate_update_8160() {
...
+  $configs = [
+    'filter.format.basic_html_plus.yml',
+  ];
+  $config_importer->importConfigs($configs);
+}
+
+/**
+ * Set format of existing job_adverts to basic_html_plus.
+ */
+function jcms_migrate_update_8163() {
```
nlisgo, line 1725: Changing the field format must be done for existing content.

sync/editor.editor.basic_html_plus.yml
```diff
@@ -0,0 +1,57 @@
+uuid: 4548e854-e092-44c4-bdf6-5165516cd0ac
```
nlisgo, line 1: This need text format was necessary because the `basic_html` which is used for impact statement does not support the `<p>` tag. We need it for these fields.

composer.json
```diff
@@ -326,7 +326,8 @@
     "drupal/duration_field": "^1.6",
     "ckeditor/codesnippet": "^4.9",
     "drupal/file_entity": "~2.0-beta6",
-    "egulias/email-validator": "^2.1"
+    "egulias/email-validator": "^2.1",
+    "drupal/allowed_formats": "^1.3"
```
nlisgo, line 330: The allowed_formats module allows us to configure a field's available text formats preventing the user from switching to a format that we don't want to support for that field. We could use this more extensively to protect fields such as impact statement which should only use very limited html.

See: https://app.zenhub.com/workspaces/journal-5ede53e8b9b11773fd13f981/issues/elifesciences/issues/6128

src/modules/jcms_migrate/jcms_migrate.install
```diff
@@ -1699,3 +1699,48 @@ function jcms_migrate_update_8160() {
...
+  \Drupal::service('module_installer')->install(['config_import']);
+}
+
+/**
+ * Update config to introduce new text format basic_html_plus.
+ */
+function jcms_migrate_update_8162() {
+  $config_importer = \Drupal::service('config_import.importer');
+  $configs = [
+    'filter.format.basic_html_plus.yml',
```
nlisgo, line 1717: This text format needs to be available to switch the format of fields on existing job adverts.

sync/filter.format.basic_html.yml
```diff
@@ -16,7 +16,7 @@ filters:
     status: true
     weight: -10
     settings:
-      allowed_html: '<i> <sub> <sup> <span class=""> <del> <math> <a href> <b> <br> <table> <caption> <tbody> <thead> <tfoot> <th> <td> <tr> <img src alt width height> <ol> <ul> <li>'
+      allowed_html: '<i> <sub> <sup> <del> <math> <a href> <b> <br> <table> <caption> <tbody> <thead> <tfoot> <th> <td> <tr> <img src alt width height> <ol> <ul> <li>'
```
nlisgo, line 19: Dropping support for `<span>` should reduce a lot of noise in our markup. This is out of scope but a minor change that will add a lot of value.

PR #685, opened 2020-10-20
Social image doi

nlisgo:


nlisgo:
I plan to deploy this to continuumtest tomorrow. We should be able to test the fragment api post and delete methods with the addition of the new social image.

PR #706, opened 2020-11-05
removes api-dummy.sha1 pin file

lsh-0:


lsh-0:
is failing because the formula is still trying to pin the version of api-dummy.

this PR is blocked by: https://github.com/elifesciences/journal-cms-formula/pull/70

PR #715, opened 2020-11-10
containerisation

lsh-0:


docker-compose.yml
```diff
@@ -0,0 +1,57 @@
+version: "3.8"
+services:
+    # basic drupal site installation with dependencies and permissions configured
+    # use `docker exec -ti journal-cms_app_1 /bin/bash` to get a shell
+    app:
+        build: .
+        image: elifesciences/journal-cms #:${IMAGE_TAG}
+        volumes:
+            # todo: ensure writeable by www-data
+            - /var/log/journal-cms:/srv/journal/private/monolog
```
nlisgo, line 10: If `/var/log/journal-cms` is supposed to be a route to a folder in the host machine I suggest that we have a folder within the project `./private/monolog` perhaps?

container/prod/configure.sh
```diff
@@ -0,0 +1,48 @@
...
+# if it looks like it's installed we can skips some steps
+rm -f site-was-installed.flag
+site_installed=false
+if ../vendor/bin/drush cget system.site name; then
+    touch site-was-installed.flag
+    site_installed=true
+fi
+
+if ! $site_installed; then
+    ../vendor/bin/drush site-install config_installer -y
```
giorgiosironi, line 18: this completes correctly for me, after checking the site was not installed

container/prod/configure.sh
```diff
@@ -0,0 +1,48 @@
...
+    redis-cli flushall
```
giorgiosironi, line 19: this errors out for me, because it talks to `127.0.0.1` whereas it should probably be talking to `journal-cms_redis_1`?

giorgiosironi: or the service name: `redis`

container/prod/configure.sh
```diff
@@ -0,0 +1,48 @@
...
+../vendor/bin/drush config-import -y
+../vendor/bin/drush cr
+
+if [[ $(../vendor/bin/drush php-eval "print node_access_needs_rebuild()") == "1" ]]; then
+    ../vendor/bin/drush php-eval "node_access_rebuild();"
+fi
+
+rm -f /tmp/drush-migrate.log
+../vendor/bin/drush mi jcms_subjects_json 2>&1 | tee --append /tmp/drush-migrate.log
+cat /tmp/drush-migrate.log | ../check-drush-migrate-output.sh
```
giorgiosironi, line 32: `check-drush-migrate-output.sh` seems not be copied over in the Dockerfile

docker-compose.yml
```diff
@@ -0,0 +1,57 @@
+version: "3.8"
+services:
+    # basic drupal site installation with dependencies and permissions configured
+    # use `docker exec -ti journal-cms_app_1 /bin/bash` to get a shell
+    app:
+        build: .
```
giorgiosironi, line 6: would be worth to put `./mysql` in a `.dockerignore` file, as it's accidentally picked up during a new `docker-compose buil` and sent to the daemon (this happen regardless of whether it's mentioned in `COPY~ statements)

docker-compose.yml
lsh-0: go for it

docker-compose.yml
lsh-0: it's permissions are wonky as well after creation which makes the recreation a bit of a pain.

I imagine a local mysql container is only there for non-prod.

container/prod/configure.sh
lsh-0: the configuration should be using the hostnames set up by docker-compose, which are `redis` and `mysql` (I think).

configuration files are in `./config/drupal-container.settings.php`

lsh-0:
wow - it's been a year.

scottaubrey:
Work has been restarted here: https://github.com/elifesciences/journal-cms/pull/1122

PR #733, opened 2020-11-20
Remove + from supported characters in image upload filename

nlisgo:


NuclearRedeye (commented):
My only comment would be that you should add a unit test for the `filenameSanitize` function ideally to confirm that it behaves as expected with a variety of filenames.

nlisgo:
> My only comment would be that you should add a unit test for the `filenameSanitize` function ideally to confirm that it behaves as expected with a variety of filenames.

I would agree. We don't have a lot of js in journal-cms but where we have it we should have tests. We don't have a single js test on this repo so I am reluctant to add it as a blocker here. Would you be ok with it as a followup? We could pick it up next sprint?

NuclearRedeye:
Yep, happy for that to be followed up. Just raise an issue and then happy for this to go in :-)

nlisgo:
> Yep, happy for that to be followed up. Just raise an issue and then happy for this to go in :-)

Issue raised: https://github.com/elifesciences/issues/issues/6292

PR #740, opened 2020-11-26
drupal-vm 6.0.2 update

nlisgo:


config/config.yml
```diff
@@ -49,7 +49,7 @@ drupalvm_vagrant_version_min: '2.2.0'
 drupalvm_ansible_version_min: '2.8'
 
 # Force use of ansible_local provisioner, even if Ansible is installed on host.
-force_ansible_local: false
+force_ansible_local: true
```
nlisgo, line 52: This gets rid of a python error that I am experiencing since the Big Sur update. I prefer the approach of running as much within the vm as possible.

composer-setup.json
```diff
@@ -1,13 +1,5 @@
 {
   "require": {
-    "geerlingguy/drupal-vm": "6.0.0",
-    "cweagans/composer-patches": "1.6.7"
-  },
-  "extra": {
-    "patches": {
-      "geerlingguy/drupal-vm": {
-        "Fix php 7.4 not purged on debian": "https://github.com/oxyc/drupal-vm/commit/cec3dd6487a7ab5f2b1d315044e512b3695070d1.patch"
```
nlisgo, line 9: This patch is no longer necessary in `6.0.2`

NuclearRedeye (approved):

PR #797, opened 2022-07-01
bumps php version in composer.json to 7.4.

lsh-0:
removes some patches that no longer apply.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -256,7 +256,7 @@
     }
   ],
   "require": {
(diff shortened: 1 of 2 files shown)


lsh-0:
`journal-cms--ci` was out of disk space because binary log files had eaten up all the space.

fix for that here: https://github.com/elifesciences/builder-base-formula/pull/355

I've otherwise cleaned it all up. Kicking ...

PR #798, opened 2022-07-04
Upgrade email-validator library

nlisgo:


config/config.yml
```diff
@@ -259,9 +259,9 @@ firewall_disable_ufw: true
 
 # PHP Configuration. Currently-supported versions: 7.2, 7.3, 7.4.
 # See version-specific notes: http://docs.drupalvm.com/en/latest/configurations/php/
-php_version: "7.2"
+php_version: "7.4"
 php_install_recommends: no
-php_memory_limit: "256M"
+php_memory_limit: "512M"
```
lsh-0, line 264: I think we run with `128M` in prod, just FYI.

PR #806, opened 2022-10-13
Add drupal core-dev - step 5

paulmartin84:


nlisgo (approved):
This looks good to me. Could you just summarise why we need the `phpcs:ignore` or `phpcs:disable` so that we can avoid those violations in future and possibly address these few uses in a followup post-Drupal9.

PR #807, opened 2022-10-16
Upgrade to Drupal 9 - step 6

paulmartin84:


nlisgo:
Failed on the composer install step:

```
[RuntimeException]
Could not delete /ext/srv/journal-cms/web/sites/default/default.services.yml: Module 'newrelic' already loaded
```

PR #809, opened 2022-10-29
Remove config installer - step 0

nlisgo:


nlisgo:
@paulmartin84 I have created step 0

PR #946, opened 2024-02-14
Add DDev installation instructions to README.md

soggy-mushroom:


nlisgo:
Thanks @soggy-mushroom. Were you able to get journal-cms running locally?

Happy with the README.md change but not sure about the drush update. We should not do that as part of this PR.

PR #956, opened 2024-03-04
Contrib module updates

Dippers:
Contrib updates for:
address
admin_toolbar
config_devel
country
ctools
devel
entityqueue
field_permissions
field_validation
hide_revision_field
inline_entity_form
migrate_plus
migrate_tools
paragraphs
redis
token
transliterate_filenames

A patch to entity_reference_revisions has been included in the module and is no longer required.


nlisgo:
@Dippers this all looks good. I may merge this in today. I am worried that it may change the target for some of the end2end tests which I will need to help to get through. Our end2end tests are written in python.

PR #976, opened 2024-04-23
Use media oembed to resolve tweet embeds in jcms_admin

Dippers:


nlisgo:
We need to enable media module and then export config as we have the following error:

```
The service "jcms_admin.tweet" has a dependency on a non-existent service "  
[2024-04-23T07:24:55.982Z] [54.226.235.133] out:                     media.oembed.url_resolver".
```

We will need to adjust tests in `src/modules/jcms_admin/tests/src/Unit/TweetTest.php` also.

nlisgo (changes requested):
See: https://github.com/elifesciences/journal-cms/pull/976#issuecomment-2071619369

